PR #3, opened 2015-07-14
add versioneer for consistent version numbers

wulmer:
The last PyPI release of pyads is 1.1.1, but `PKG-INFO` shows 1.0.2. Documentation shows 1.1.0.

This PR adds versioneer to the pyads package which should simplify version handling. Simply tag a commit and then run `setup.py build|sdist` or `make html` to create a package or the documentation using the version given by the git tag.

More details on versioneer can be found [here](https://github.com/warner/python-versioneer)

BTW: I would not add `/doc/build/` to the repo, it is usually ignored.


MrLeeh:
Thanks for the PR. Seems like a good feature. Although I thought for such a small project it would be easy to keep version numbers consistent. As one can see it doesn' t. What I don' t like about it is that it adds a lot of code just for version handling.

wulmer:
True. This PR includes a lot of code. But when you look closer, the largest parts are:
* `versioneer.py` (which is just a script used during development, will not be added to releases)
* `pyads/_version.py` (will be replaced by a static version during deployment, try `python setup.py build` and check `build/lib/pyads/_version.py`).

And the rest just simplifies spread version numbers by a single central one.

MrLeeh:
I get the point. I will have a look at it.

MrLeeh:
Allright, I played a bit with versioneer and it is actually a pretty nice piece of code. So I included it in 1.1.2.

PR #5, opened 2015-07-23
Use windll, and add adsSyncReadByName(), adsSyncWriteByName()

mdauphin:
To work on my x64 windows with python x32, I must use ctypes.windll to load TcAdsDll.
Add two function to access data by variable name:
adsSyncReadByName() and adsSyncWriteByName()
ps: I newbe in github I dont know if is was the good way


MrLeeh:
Alright, I just tested your code with the windll version. Seems to work on 32 Bit systems and 64 bit systems. Don' t know anymore why I used the CDLL version...Maybe it was a Windows XP thing.
I will test your Name functions as well and then merge your Pull request.

PR #20, opened 2017-07-13
Notification callback on variable changes

pjaneck:
Hey Stefan,
I think this is easier than applying a patch with no comments.
Regards,
Peter


MrLeeh:
I will create a new branch dev until we got proper testing for this.

xboomx:
Hi everyone, 
maybe my scenario is not so typical... but could you add another method like "add_device_notification" for use in case you don't have variable name - by address? I have only indexGroup and indexOffset for %MW variable.

MrLeeh:
I opened an issue for this: #39

PR #32, opened 2017-09-07
enhancement: passing an integer to the callback

pjaneck:
Hey Stefan,
this feature from the device notification hasn't been implemented jet. Unitl now the call would pass the handle as default value. However, as I understand it, you can pass any value:
hUser
    [in] 32-bit value that is passed to the callback function.
Regards,
Peter

### DIFF
--- a/pyads/ads.py
+++ b/pyads/ads.py
@@ -254,12 +254,12 @@ def delete_route(adr):
     return adsDelRoute(adr.netIdStruct())
 
 
-def add_device_notification(adr, data_name, attr, callback):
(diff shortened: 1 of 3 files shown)


MrLeeh:
Hey Peter, looks OK to me. I will accept the Pull Request but change the name in a more python *user_handle*.

pjaneck:
Hey Stefan,
could you increase the version number and upload the new version to pypi?
One more request, could you create a wheel package for python 2 (yes, there are still people using it)?
Thanks in advance,
Peter

PR #45, opened 2017-12-01
Fix for issue #39

pjaneck:
The following two commits should fix issue MrLeeh/pyads#39

Regards,
Peter


stlehmann:
This PR ist handled by #125

PR #48, opened 2018-03-10
added PLCTYPE_LINT and PLCTYPE_ULINT to plc data types.

ASolchen:
According to the Beckhoff website the ULINT and LINT are not supported by TwinCAT. I believe this information is outdated because TwinCAT3 programming software does have these types. I ran the following as a test with success. 
`import pyads
from ctypes import c_uint64, c_int64
plc = pyads.Connection('127.0.0.1.1.1',851)
print(plc)
if plc:
    plc.open()
    val = plc.read_by_name('Global.test_lint', c_int64)
    print(val)
    val = plc.read_by_name('Global.test_ulint', c_uint64)
    print(val)`

### DIFF
--- a/pyads/constants.py
+++ b/pyads/constants.py
@@ -10,7 +10,7 @@
 
 """
 from ctypes import c_bool, c_byte, c_int8, c_uint8, c_int16, c_uint16, \
-    c_int32, c_uint32, c_float, c_double, c_char, c_short
+    c_int32, c_uint32, c_float, c_double, c_char, c_short, c_int64, c_uint64
 
 STRING_BUFFER = 1024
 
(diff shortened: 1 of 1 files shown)


stlehmann:
@ASolchen Did you try this with TwinCAT2 and TwinCAT3?

ASolchen:
I installed TwinCAT2 into a VM and there are no atomic data types of LINT or ULINT. I still believe these could be added as it does not make to library unstable. A user of TwinCAT2 could make a data type LINT or ULINT by making an array of BYTE[8], WORD[4], or DINT[2] and packing it accordingly. The library should still work.

PR #58, opened 2018-11-07
Change notification decorator to process ctype.Structure objects

briggs360:
The following commit allows the ability to receive structures in notifications by modifying the decorator.

For example:

```
class TowerEvent(Structure):
    _fields_ = [
        ("Category", c_char * 21),
        ("Name", c_char * 81),
        ("Message", c_char * 81)
    ]

@plc.notification(TowerEvent)
def callback(handle, name, timestamp, value):
    print(f'Received new event notification for {name}.Message = {value.Message}')
```
^^^ Sample code modified to work with the latest commit.

Regards,
David

### DIFF
--- a/pyads/ads.py
+++ b/pyads/ads.py
@@ -11,7 +11,7 @@
 from typing import Optional, Union, Tuple, Any, Type, Callable, Dict
 from datetime import datetime
 import struct
-from ctypes import memmove, addressof, c_ubyte
+from ctypes import memmove, addressof, c_ubyte, Structure
 
 from .utils import platform_is_linux
 from .filetimes import filetime_to_dt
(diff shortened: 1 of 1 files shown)


dabrowne:
An issue I can see here is that you're creating a shared instance of your `Structure` subclass and mutating it on each notification. In the event of multiple notifications, the shared instance will be mutated each time, which becomes an issue if it is persisted somewhere.

I'd use the following instead:

```python
elif issubclass(plc_datatype, Structure):
    value = plc_datatype()
    memmove(addressof(value), addressof(data), data_size)
```

The decorator would then be used like this (passing TowerEvent class rather than instance):

```python
@plc.notification(TowerEvent)
def callback(handle, name, timestamp, value):
    print(f'Received new event notification for {name}.Message = {value.Message}')
```

dabrowne:
We should also be a bit more careful with the memmove and use something like this to avoid overwriting adjacent memory in case we get more data then expected*:

```python
from ctypes import sizeof

elif issubclass(plc_datatype, Structure):
    value = plc_datatype()
    fit_size = min(data_size, sizeof(value))
    memmove(addressof(value), addressof(data), fit_size)
```

*[reference](https://wiki.python.org/moin/ctypes)

briggs360:
Thanks for the suggestions @dabrowne. Verified working with my test system.

stlehmann:
Thanks for the PR and thanks to Dan for the review. It looks fine to me now, so I will accept the PR.

PR #62, opened 2019-03-25
Update pyads_ex.py

### LINKED ISSUE #61: array datatypes for function read_write
Hi,
I am currently trying to use the function `read_write` on a `Connection` instance. However I figured out that for the write parameter (`plc_write_datatype`) only none-array types are working. It would be very nice to be able to use array types as well.
I assume implementing this is quiet trivial as it is already working fine for the `write` function.
Just copying the folllowing code from `pyads_ex.py` / `adsSyncWriteReqEx` 
```python
    else:
        if type(plc_data_type).__name__ == "PyCArrayType":
            data = plc_data_type(*value)
        else:
            data = plc_data_type(value)
```
.. to `adsSyncReadWriteReqEx2`. But changing `plc_data_type` to `write_data_type`.

PS: Thanks for sharing your great work on this Python package.

bpartzsch:
I want to resolve #61. And allow users to use array types for write data on function `readwrite`.

### DIFF
--- a/pyads/pyads_ex.py
+++ b/pyads/pyads_ex.py
@@ -436,7 +436,10 @@ def adsSyncReadWriteReqEx2(
         # Add an extra byte to the data length for the null terminator
         write_length = len(value) + 1
     else:
-        write_data = write_data_type(value)
+        if type(write_data_type).__name__ == "PyCArrayType":
+            write_data = write_data_type(*value)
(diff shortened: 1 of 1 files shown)


stlehmann:
Thanks for the PR. You can find your changes in the new version 3.0.8 on PyPi.

Issue #64, opened 2019-03-30
"struct.error: unpack requires a buffer of 1 bytes" when trying to add notification


EsbenLorenzen:
I have a Twincat 2 project, with UDT's as global variabels.

I was hoping to be able to get notifications when boolean-variables changes, by doing the following:

`plc.connection.add_device_notification('.dStatus.bHardwareOk', pyads.NotificationAttrib(4), cb)`

My callback function looks like this:
```python
@plc.connection.notification(pyads.PLCTYPE_BOOL)
    def callback(handle, name, timestamp, value):
        print('{}: recieved notification for {} :: {}'.format(timestamp, name, value))
```

I get the following error:

> Traceback (most recent call last):
>   File "_ctypes/callbacks.c", line 232, in 'calling callback function'
>   File "C:\Users\Esben L\AppData\Local\Programs\Python\Python37-32\lib\site-packages\pyads\pyads_ex.py", line 678, in wrapper
>     return callback(notification, data_name)
>   File "C:\Users\Esben L\AppData\Local\Programs\Python\Python37-32\lib\site-packages\pyads\ads.py", line 737, in func_wrapper
>     bytearray(data)[:data_size]
> struct.error: unpack requires a buffer of 1 bytes

When I add a variable directly, like "MAIN.testbool", it works as expected. Reading the README, I see that the callback notification is not supported for structs.

Is there any way to get around this? Almost all of my variables are places in structs, for.... well for structures sake ;)

Any help and tips would be appreciated! And thank you for the effort put into this project!

EsbenLorenzen:
I think I have solved this by myself by creating a NotificationAttribute "by hand":

```python
attr = pyads.NotificationAttrib(sizeof(pyads.PLCTYPE_BOOL), trans_mode=4)
```

An using that when adding my notification:
```python
cb_handles.append(plc.connection.add_device_notification('.dStatus.bHardwareOk', attr, cb))
```

This seems to be working. I guess I was passing the wrong size to the wrapper.

Now in regards to notification handles and how they are disposed, is there a way to "clear" all the added notification?

Or is it a matter of saving the user_handles and deleting them one by one, before stopping the plc connection?
In production the notification setup would only be done once per restart of the machine, however developing, I am constantly reloading the python-script, while the PLC keeps going. Is this causing any "clutter" on the PLC/AMS Router?

pyhannes:
I'm facing the same issue as you but.
[...]

PR #68, opened 2019-05-28
Fix SAdsNotificationHeader structure

### LINKED ISSUE #43: Notification limited to 8 byte?
I am trying to add a notification for a _STRUCT_ that contain 571 _bool_

When the notification is received I can only access 8 bytes while the datasize is 598 (should be 571?): 
```
In [5]: contents.cbSampleSize
Out[5]: 598
In [6]: len(bytearray(contents.data))
Out[6]: 8
```

From wireshark I can see the following:
image

This is on Windows...

dxpke:
The data field is a dynamically sized array that is part of the
structure. As such, declaring it as a pointer type is wrong.
If it is declared as a pointer, `bytearray(data)` is limited in size
to the size of a pointer on the platform Python is running on.
With x86 (32-bit) Python, it is not possible to capture double-
precision floating point values this way, since they are twice
as large as the pointer size. `struct.unpack` will then complain
that the sizes do not match up.
Instead, declare that the data field is one byte (like in the actual
C structure), and dynamically build a field of the right size
when it is known.
This additionally allows to remove the superfluous copy of
the value for string parsing.

This breaks backwards-compatibility when functions relied on
getting a pointer-sized byte array for unmapped types since
they are now also correctly sized.

Fixes #43

### DIFF
--- a/pyads/ads.py
+++ b/pyads/ads.py
@@ -57,7 +57,7 @@ from .constants import (
     PLCTYPE_WORD,
 )
 
-from .structs import AmsAddr, SAmsNetId, AdsVersion, NotificationAttrib
+from .structs import AmsAddr, SAmsNetId, AdsVersion, NotificationAttrib, SAdsNotificationHeader
 
 linux = platform_is_linux()
 port = None  # type: int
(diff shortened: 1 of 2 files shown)


stlehmann:
Thanks for the PR. As I understand it this might also fix issue https://github.com/stlehmann/pyads/issues/64?

dxpke:
I think so, yes.

stlehmann:
Thanks again for your help with this bug. There should definetely be a test for this use case of Device Notifications with larger data types. I want to ask you if you would like to add one. I would add it myself but I am not doing anything with pyads at the moment so I would need to find the time to get into it again. Kind Regards.

dxpke:
Test case added in #111

PR #73, opened 2019-07-11
FIX: pyads.structs structure definitions

klauer:
Details in issue #72

Closes #72

PR #74, opened 2019-07-11
Query server for data type if not specified by user

klauer:
The overall idea is that the server/PLC can be queried for type information for a given symbol by name, and - in most of our use cases, at least - it's unnecessary for the user to specify that expected type ahead of time. Python's introspection abilities are more than sufficient to determine what to do with the value when received.

Opening this as a draft for your consideration. If we can agree this is a worthwhile addition, I'd be happy to change things around to better match the API you'd prefer.

Here's a few notes:
1. PR #73 is a prerequisite, as the structure definition does not allow the type name to be accessed as-is
    a. There's a bit of a change in `adsSyncReadWriteReqEx2`, as the expected length read back is wholly dependent on the length of the name, type, and comment strings held in `SAdsSymbolEntry`
2. ADS data types constants were added from https://infosys.beckhoff.com/english.php?content=../content/1033/tcplclib_tc2_utilities/9007199290071051.html&id=
3. Type map dictionary from `ADST_*` to ctypes type were added
    a. The server responds with both the integer `ADST_*` constant, and also the type name as a string
    b. With the above, there's the possibility to make this a user-hook, allowing either matches from the `ADST_` type ID or the type name string.
3. Repeatedly querying the type should be unnecessary - perhaps a connection-level cache could be used to store type information. 
4. Writes are not yet handled, but a similar method could be used there. Haven't looked into the notification API.
5. The test server will need some modifications to actually get this tested in the suite. I have tested it interactively on an OSX machine connecting to a CX5010, in any case.

cc @zllentz @teddyrendahl


stlehmann:
I need to have a closer look at this but I definetely like the idea. I didn't even know ADS supported a call for type information, until now.  So for a start pimping the functions read(), read_by_name(), write() and write_by_name() would be great.

klauer:
Thanks for the feedback, @stlehmann. 👍 I'll look further into fully integrating the concept.

stlehmann:
@klauer Long time no see :) Did you make any further progress on this issue?

klauer:
Hi @stlehmann, sadly I haven't gotten around to this just yet. I really hope to at some point, though. I don't mind if this gets closed in the meantime.

klauer:
@bdcoyle is absorbing my work here into his PR. Closing.

PR #75, opened 2019-07-12
Linux adding routes to remote PLC

Adrian-at-CrimsonAuzre:
It's a little bare bones, doesn't have tests, and could probably do with some more work, but it successfully adds a route to the remote PLC on demand. I made these modifications for a project I was working on and I figured I'd share.

A rough breakdown of the data structure:
image

Would love some feedback.

### DIFF
--- a/pyads/pyads_ex.py
+++ b/pyads/pyads_ex.py
@@ -34,7 +34,7 @@ from .constants import (
     PLCTYPE_UDINT,
     ADSIGRP_SYM_VALBYHND,
     ADSIGRP_SYM_RELEASEHND,
-	PORT_SYSTEMSERVICE
+    PORT_SYSTEMSERVICE
 )
 from .errorcodes import ERROR_CODES
 
@@ -165,6 +165,8 @@ def adsAddRouteToPLC(sending_net_id, ip_address, username, password, route_name=
 
     """
     import socket
+    import struct
+    from contextlib import closing
     # ALL SENT STRINGS MUST BE NULL TERMINATED
     adding_host_name = adding_host_name + '\0' if adding_host_name else socket.gethostname() + '\0'
     added_net_id = added_net_id if added_net_id else sending_net_id
@@ -174,29 +176,29 @@ def adsAddRouteToPLC(sending_net_id, ip_address, username, password, route_name=
     password = password + '\0'
 
     # The head of the UDP AMS packet containing host routing information
-    data_header = (0x036614710000000006000000).to_bytes(12, 'big')
+    data_header = struct.pack('>12s', b'\x03\x66\x14\x71\x00\x00\x00\x00\x06\x00\x00\x00')
     data_header += bytes(map(int, sending_net_id.split('.')))		# Sending net ID
-    data_header += PORT_SYSTEMSERVICE.to_bytes(2, 'little')			# Internal communication port
-    data_header += (0x0500).to_bytes(2, 'big')						# Write command
-    data_header += (0x00000c00).to_bytes(4, 'big')					# Block of unknown
-    data_header += len(adding_host_name).to_bytes(2, 'little')		# Length of sender host name
-    data_header += adding_host_name.encode('utf-8')				# Sender host name
-    data_header += (0x0700).to_bytes(2, 'big')						# Block of unknown
+    data_header += struct.pack('<H', PORT_SYSTEMSERVICE)			# Internal communication port
+    data_header += struct.pack('>2s', b'\x05\x00')					# Write command
+    data_header += struct.pack('>4s', b'\x00\x00\x0c\x00')			# Block of unknown
+    data_header += struct.pack('<H', len(adding_host_name))			# Length of sender host name
+    data_header += adding_host_name.encode('utf-8')					# Sender host name
+    data_header += struct.pack('>2s', b'\x07\x00')					# Block of unknown
 
 
-    actual_data = (6).to_bytes(2, 'little')							# Byte length of AMS ID (always 6)
+    actual_data = struct.pack('<H', 6)								# Byte length of AMS ID (always 6)
     actual_data += bytes(map(int, added_net_id.split('.')))			# Net ID being added to the PLC
-    actual_data += (0x0d00).to_bytes(2, 'big')						# Block of unknown (maybe encryption?)
-    actual_data += len(username).to_bytes(2, 'little')				# Length of the user name field
+    actual_data += struct.pack('>2s', b'\x0d\x00')					# Block of unknown (maybe encryption?)
+    actual_data += struct.pack('<H', len(username))					# Length of the user name field
     actual_data += username.encode('utf-8')							# PLC Username
-    actual_data += (0x0200).to_bytes(2, 'big')						# Block of unknown
-    actual_data += len(password).to_bytes(2, 'little')				# Length of password field
+    actual_data += struct.pack('>2s', b'\x02\x00')					# Block of unknown
+    actual_data += struct.pack('<H', len(password))					# Length of password field
     actual_data += password.encode('utf-8')							# PLC Password
-    actual_data += (0x0500).to_bytes(2, 'big')						# Block of unknown
-    actual_data += len(route_name).to_bytes(2, 'little')			# Length of route name
+    actual_data += struct.pack('>2s', b'\x05\x00')					# Block of unknown
+    actual_data += struct.pack('<H', len(route_name))				# Length of route name
     actual_data += route_name.encode('utf-8')						# Name of route being added to the PLC
 
-    with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as sock: # UDP
+    with closing(socket.socket(socket.AF_INET, socket.SOCK_DGRAM)) as sock: # UDP
         # Listen on 55189 for the response from the PLC
         sock.bind(('', 55189))
 
@@ -223,9 +225,9 @@ def adsAddRouteToPLC(sending_net_id, ip_address, username, password, route_name=
         rcvd_protocol_block = data[22:]						# Unknown block of protocol
         rcvd_is_password_correct = rcvd_protocol_block[4:7]	# 0x040000 when password was correct, 0x000407 when it was incorrect
    
-    if int.from_bytes(rcvd_is_password_correct, 'big') == 0x040000:
+    if rcvd_is_password_correct == b'\x04\x00\x00':
         return True
-    elif int.from_bytes(rcvd_is_password_correct, 'big') == 0x000407:
+    elif rcvd_is_password_correct == b'\x00\x04\x07':
         return False
     else:
         raise ValueError('Received unknown response from ' + ip_address)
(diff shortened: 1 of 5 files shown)


Adrian-at-CrimsonAuzre:
Little piece of code I used to test it against a PLC
```
import pyads

print(pyads.__file__)
my_ads_address = '169.254.104.190.1.1'
target_ip = '192.168.200.210'
target_ads = '5.6.182.28.1.1'
if pyads.add_route_to_plc(my_ads_address,target_ip,'administrator','1'):
	# test PLC connection
	pyads.set_local_address(my_ads_address)
	with pyads.Connection(target_ads, pyads.PORT_SPS1, target_ip) as plc:
		temp = plc.read_by_name('.PARK_ROBOT', pyads.PLCTYPE_BOOL)
		print(temp)

```

In therory, this can also be used on Windows but there isn't really a reason to use it because you can't add routes on the client side.

stlehmann:
First, thank you for this PR. This looks like a handy extension and I think it is a good idea to add it. However, I don' t want to add any new untested code to the repo so I need to ask you to also implement some test-code for this function. Please have a look at the tests directory.

Thanks again for the effort and sorry for asking for more but it is necessary that any new features gets its automated test not only to ensure it works as described but also to make sure it doesn't fail due to futures changes.

stlehmann:
And of course we will also need some documentation for it ;)

pyads/pyads_ex.py
```diff
@@ -149,6 +149,85 @@ def adsAddRoute(net_id, ip_address):
...
+    sending_host_name = sending_host_name + '\0' if sending_host_name else socket.gethostname() + '\0'
+    added_net_id = added_net_id if added_net_id else sending_net_id
+    route_name = route_name + '\0' if route_name else sending_host_name
+
+    username = username + '\0'
+    password = password + '\0'
+
+    # The head of the UDP AMS packet containing host routing information
+    data_header = (0x036614710000000006000000).to_bytes(12, 'big')
+    data_header +=	bytes(map(int, sending_net_id.split('.')))		# Sending net ID
```
kryskool: Only one space between operator

pyads/pyads_ex.py
```diff
@@ -149,6 +149,85 @@ def adsAddRoute(net_id, ip_address):
...
+    data_header += len(sending_host_name).to_bytes(2, 'little')		# Length of sender host name
+    data_header += sending_host_name.encode('utf-8')				# Sender host name
+    data_header += (0x0700).to_bytes(2, 'big')						# Block of unknown
+
+
+    actual_data = (6).to_bytes(2, 'little')							# Byte length of AMS ID (always 6)
+    actual_data += bytes(map(int, added_net_id.split('.')))			# Net ID being added to the PLC
+    actual_data += (0x0d00).to_bytes(2, 'big')						# Block of unknown (maybe encryption?)
+    actual_data += len(username).to_bytes(2, 'little')				# Length of the user name field
+    actual_data +=	username.encode('utf-8')						# PLC Username
```
kryskool: Only one space between operator

pyads/pyads_ex.py
```diff
@@ -149,6 +149,85 @@ def adsAddRoute(net_id, ip_address):
...
+        # Convert to a String AMS ID
+        #rcvd_PLC_AMS_ID = '.'.join([str(int.from_bytes(rcvd_PLC_AMS_ID[i:i+1], 'big')) for i in range(0, len(rcvd_PLC_AMS_ID), 1)])
+        rcvd_AMS_port = data[18:20]							# Some sort of AMS port? Little endian
+        rcvd_command_code = data[20:22]						# Command code (should be read) Little endian
+        rcvd_protocol_block = data[22:]						# Unknown block of protocol
+        rcvd_is_password_correct = rcvd_protocol_block[4:7]	# 0x040000 when password was correct, 0x000407 when it was incorrect
+   
+    if int.from_bytes(rcvd_is_password_correct, 'big') == 0x040000:
+        return True
+    else:
```
kryskool: If value is 0x040000 when password was correct and 0x000407 when it was incorrect, we must test both
and return an exception if none of this values are read

pyads/pyads_ex.py
Adrian-at-CrimsonAuzre: a064e61d4c939ce2cdc19e315291eab87c945890

pyads/pyads_ex.py
Adrian-at-CrimsonAuzre: a064e61d4c939ce2cdc19e315291eab87c945890

pyads/pyads_ex.py
Adrian-at-CrimsonAuzre: a34d4d596f185f0e5a26aa8b842b4c5c74403333

Adrian-at-CrimsonAuzre:
I'll work on some tests later today, running into issues connecting to one of our Beckhoff PLCs on Linux. Anyone have any input? #77

Adrian-at-CrimsonAuzre:
[Ah, Python 2.](https://travis-ci.org/stlehmann/pyads/jobs/559526512) Wish I would have checked that earlier...

I don't really have experience using Python 2 and I personally don't want to rewrite this PR to support a version that is EOL in [5 months](https://pythonclock.org/).

stlehmann:
>  I personally don't want to rewrite this PR to support a version that is EOL in 5 months.

I understand that. Maybe it's about time to drop the Python2 support in the future. However, as the whole codebase is Python2 compatible I would like to keep it that way for now. I will have a closer look at your PR later on and try to recover Python2 compatibility.

kryskool:
Hi @Adrian-at-CrimsonAuzre 

you can use [closing() on contextlib](https://docs.python.org/2/library/contextlib.html#contextlib.closing)

replace

```python
with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as sock:
```

like this

```python
from contextlib import closing

with closing(socket.socket(socket.AF_INET, socket.SOCK_DGRAM)) as sock:
```

regards

Adrian-at-CrimsonAuzre:
It's not just that, I make heavy use of int.to_bytes() and int.from_bytes() which (I'm assuming) would need to become struct.pack() and struct.unpack()

stlehmann:
Yes, struct.pack() is the way to go then. It is also faster then int.to_bytes(), though ;)

Adrian-at-CrimsonAuzre:
I've done what I can, but my dev environment is only set up for Python 3 so I can't go line by line to figure out what's wrong. I've just been guessing and running the tests in Python 2 so far.

pyads/pyads_ex.py
```diff
@@ -149,6 +150,87 @@ def adsAddRoute(net_id, ip_address):
...
+        # Allow TimeoutError to be raised so user can handle it how they please
+
+        # Keep looping until we get a response from our PLC
+        addr = [0]
+        while addr[0] != ip_address:
+            data, addr = sock.recvfrom(32) # PLC response is 32 bytes long
+
+    rcvd_packet_header = data[0:12]							# AMS Packet header, seems to define communication type
+    # If the last byte in the header is 0x80, then this is a response to our request
+    if rcvd_packet_header[-1] == 0x80:
```
klauer: If `rcvd_packet_header[-1] != 0x80` then `rcvd_is_password_correct` will never be defined. It'll raise `UnboundLocalError` then.

Adrian-at-CrimsonAuzre: I'm not sure what it should do here. If the packet header doesn't end in 0x80 but it did come from the PLC on the port we were listening to it would mean we got a message we shouldn't have. Maybe intentionally raising an exception would be best?

klauer:
Thanks, @Adrian-at-CrimsonAuzre - I'm rather excited about this functionality! It can be frustrating to always need a machine with TwinCAT just to add a route so ADS will work.

Bizarrely, I gave this PR a try and found that regardless of the password (correct or incorrect), `rcvd_is_password_correct == b'\x04\x00\x00'` and `add_route_to_plc` returns `True`.

Despite that positive response, I wasn't able to access the PLC over ADS from the machine I added a route to.

Then, instead of leaving `adding_host_name` to the default (`gethostname`), I specified my machine's IP address. I was then able to access the PLC over ADS. 

I wonder if `adding_host_name` should be required?

pyads/pyads_ex.py
klauer: I can't speak for the repository maintainer, but I think raising a `RuntimeError("Unexpected response from PLC")` would not be unreasonable...

Adrian-at-CrimsonAuzre:
@klauer  The route being added successfully just means that the PLC added the route, not that the route is correct. It's strange that you were getting a success response with an incorrect password though, any chance you can post the bytes returned by the PLC when this occurs?

Making it mandatory might be a good idea, I also had an issue with a Docker container that was only resolved when I specified the server hostname. It could also be an issue with the PLC being unable to resolve the hostname, in which case sending the destination IP is probably the better solution. I think TwinCAT uses the hostname because it's not a given that your computer has a static IP.

What I've learned over the course of my project is that PLCs are weird, and how they communicate is weirder.

klauer:
@Adrian-at-CrimsonAuzre Agreed on the overall weirdness. The incorrect password not making a difference in the response is the main thing that struck me as concerning, not so much the required IP vs hostname.

See the 'sent'/'received' lines in this gist if you want more info, along with the comment above each execution: https://gist.github.com/klauer/d1154d9e7a8a16bd398cec2bfc0a9678 Not sure there's much that can be done here. I'm glad it works, regardless.

Adrian-at-CrimsonAuzre:
@klauer On the PLCs I tried it returned a different response, but (most) PLCs don't seem to be designed with network security in mind. Maybe this is an artifact of an older version of TwinCAT installed on the PLC? Or maybe this is something you should report to Beckhoff?

stlehmann:
@Adrian-at-CrimsonAuzre  Has there been any progress in this matter recently? I've just checked that the tests are not passing correctly for Python2 and Python3.5. Also there seem to be some problems with connecting to certain PLCs.

Adrian-at-CrimsonAuzre:
@stlehmann No idea why the Python3.5 test failed, it passed on my local machine before I committed the last changes. Looks like it timed out when running online.
I'm personally not going to keep working on the Python2 implementation because I don't really have a proper dev environment for it and I don't have the time. I can, however, throw some documentation together if you'd like.
It seems that the connection issues stem from PLCs being strange, but presently it works for all of the Beckhoff PLCs I'm connecting to. Samba should probably be listed as a requirement for pyads on Linux because otherwise certain PLCs will be unable to respond to _any_ requests from pyads without it. I'm pretty sure it has to do with NetBIOS.

Adrian-at-CrimsonAuzre:
_Now_ Python3.5 passes. That's strange.

kryskool:
Hi @Adrian-at-CrimsonAuzre 

I thinks for Python 2.7 we need to return an exception NotImplemented and catch it on test

It's a good reason to migrate on Python 3.x if we want this functionnality

Comments ?

regards,

Adrian-at-CrimsonAuzre:
@kryskool I'm inclined to agree unless someone else wants to take up the mantle and figure out what's wrong with the Python2 implementation.

stlehmann:
I fixed the last Python2 errors and merged the PR. There were just some `bytes(` statements left that broke Python2 compatibility. Could you both please test the final implementation because I am running on Windows. If everything is nice I will create a new release.

Adrian-at-CrimsonAuzre:
Woops, guess I missed some. Thanks for taking care of that.
I out adding routes and reading tags on Python3.6 and everything seems to work.

stlehmann:
Thanks again for bringing this up and elaborating. I think that's a great new feature.

PR #79, opened 2019-07-18
FIX: ADSCOMMAND_READSTATE int32 -> uint16

### LINKED ISSUE #78: ADSCOMMAND_READSTATE data types incorrect (int32 -> uint16)
Reference: https://github.com/stlehmann/ADS/blob/b56d011f4c48d2d6033a59f4e6064edc6c7a49b6/AdsLib/AdsLib.h#L112

Merging in an updated [ADS library](https://github.com/stlehmann/ADS/pull/3) exposed the issue, causing the test suite to fail on `test_read_state` with `pyads.pyads_ex.ADSError: ADSError: parameter size not correct (1797).`

Both the server and client are expecting a `ctypes.c_int` (32-bit signed int), whereas the definition above indicates it's a 16-bit unsigned int.

klauer:
Closes #78 (further details in the issue)

### DIFF
--- a/pyads/pyads_ex.py
+++ b/pyads/pyads_ex.py
@@ -243,11 +243,11 @@ def adsSyncReadStateReqEx(port, address):
     ams_address_pointer = ctypes.pointer(address.amsAddrStruct())
 
     # Current ADS status and corresponding pointer
-    ads_state = ctypes.c_int()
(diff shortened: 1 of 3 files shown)


stlehmann:
Thanks for the PR. Merged. I also updated the adslib to the current Beckhoff master.

Issue #67, opened 2019-04-21
Add additional information to datatypes in the docu


stlehmann:
I repeatedly get requests on how to use structs and arrays in pyads. To give instant answer an extra chapter in the docu could be helpful. Contents:

* basic datatypes
* arrays
* structs

PR #83, opened 2019-08-29
additional section for extra explanation of arrays and structures

chrisbeardy:
I've added some more explanation in how to use arrays and structures to solve issue #67 , I hope this is what you were looking for?


stlehmann:
Thanks for the PR :)

PR #84, opened 2019-08-30
Improve platform support

bejer:


stlehmann:
Thanks for the PR. That was a nasty bug with platform_is_linux not using a function call. Thanks again.

Issue #82, opened 2019-08-28
Add support for reading and writing structures of multiple types


chrisbeardy:
I believe it would be beneficial to be able to read and write to structures made up of multiple types, as I understand this is not possible and it can only be done of structures made of multiples of a single type.

The way I imagine this would be a call looking something like:
```python
plc.read_by_name('MAIN.TestStruct', [pyads.PLCTYPE_LREAL,
                                     pyads.PLCTYPE_BOOL,
                                     pyads.PLCTYPE_INT])
```
This would return a list of [1.2, TRUE, 3] for example.

Alternatively new functions specifically for reading/writing STRUCT types. 

Am happy to look into to this but was not sure the best way to proceed,looking at my example this would require changes all the way down to adsSyncReadReqEx2 etc to change the data_type parameter to type LIST,  then would require some manipulation to get al the returns. Not sure you would want to change the input type just to add this feature. 

Can anyone think of a better way to implement this why maintaining compatibility?

I think this is benefical as it allows just a single read to read all variables , as it would be quicker on the python code and less intensive on the ADS router. For example the PLC at the end of the cycle copies all required variables into one "super" STRUCT which can then be read more frequently. I have used the VB.NET examples from Beckhoff in the past to read such a structure.

Doing this also to callback notifications would then also be very useful!

stlehmann:
@chrisbeardy Thanks for sharing your thoughts and elaborating it in PR #88. I need to look over the implementation but I really like the idea. The feature of reading/writing structs with elements of multiple datatypes was missing so far. I would like to add this feature to the code base. Of course we will also need testing for it. So if you have already done some testing maybe you would like to add the test-code to the PR.

chrisbeardy:
@stlehmann no formal testing yet, I'll work on that shortly and add it in, just wanted to get a draft in to get your thoughts on the implementation.

chrisbeardy:
new draft pull request #92

stlehmann:
OK, I just gave it a try. I used a structure definition like so:

```python
mystruct = (
    ("b", pyads.PLCTYPE_BOOL, 1),
    ("i", pyads.PLCTYPE_INT, 1),
)
```

Turns out that the bool takes two bytes of memory. ;)

chrisbeardy:
Did you define your structure in twincat with the 1 byte alignment pack mode? 

https://infosys.beckhoff.com/content/1033/tc3_plc_intro/2529746059.html?id=3686945105176987925

[...]

PR #88, opened 2019-09-13
added function 'read_structure_by_name'

chrisbeardy:
I've added a method inside the connection class the for reading structures containing multiple types, (part way to solving #82 ), you have to pre define in python your structure definition, then it essentially calculates how many bytes to request, then converts these bytes to a list of all the numbers using the defined structure. 

Let me know what you think.

### DIFF
--- a/doc/quickstart.md
+++ b/doc/quickstart.md
@@ -208,7 +208,7 @@ Then you can read the structure:
 
 ```python
 >>> plc.read_structure_by_name('global.sample_structure', structure_def)
-[11.1, 22.2, 3, 4, 44, 444, 'abc']
+OrderedDict([('rVar': 11.1), ('rVar2': 22.2), ('iVar': 3), ('iVar2': [4, 44, 444]), ('sVar': 'abc')])
 ```
 
 ### Read and write values by address
--- a/doc/quickstart.md
+++ b/doc/quickstart.md
@@ -126,13 +126,10 @@ For reading strings the maximum buffer length is 1024.
 'abc'
 ```
 
-### Read and write arrays or structures by name
+### Read and write arrays by name
 
-You can also read/write arrays and structures. For this you simply need to multiply the datatype by
-the number of elements in the array or structure you want to read/write. Only structures made up of
-the same datatype can be used.
-
-**Arrays**
+You can also read/write arrays. For this you simply need to multiply the datatype by
+the number of elements in the array or structure you want to read/write. 
 
 ```python
 >>> plc.write_by_name('global.sample_array', [1, 2, 3], pyads.PLCTYPE_INT * 3)
@@ -146,7 +143,9 @@ the same datatype can be used.
 5
 ```
 
-**Structures**
+### Read and write structures by name
+
+#### Structures of the same datatype
 
 TwinCAT declaration:
 ```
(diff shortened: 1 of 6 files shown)


doc/quickstart.md
```diff
@@ -175,6 +174,43 @@ Python code:
...
+...    ('iVar', pyads.PLCTYPE_INT, 1),
+...    ('iVar2', pyads.PLCTYPE_DINT, 3),
+...    ('sVar', pyads.PLCTYPE_STRING, 1))
+```
+
+Then you can read the structure:
+
+```python
+>>> plc.read_structure_by_name('global.sample_structure', structure_def)
+[11.1, 22.2, 3, 4, 44, 444, 'abc']
```
stlehmann: As it is named read_structure_by_name I think it would make more sense to return a dictionary or a named tuple with the defined member names. What do you think?

chrisbeardy: My initial instinct when I first starting thinking of this was to return a dictionary with the key names being the variable names as defined in the 'structure_def' as it does seem more intuiative for a structure. However I decided for the draft to go with a list to keep the return consistent with the return forms of the other functions.

I also thought that if I create a 'write_structure_by_name' method it would be easier to pass a list in to write, rather than try to interpret a dictionary?

I'm not sure as a user if they would prefer to deal with lists and then use the indexes of the list (as they know the structure definition) to get the variable or if they would prefer to deal with dictionary. Which one do you think would be easier for users to deal with in their program? I guess a dictionary makes sense if you were to not use all the variables that are in the structure as you can just pick the ones you want?

I am also still working on further tests to keep the coverage up and shall also aim to implement the ability to optionally pass the handle in like my other PR and also I believe it makes sense to be able to optionally pass the structure size in too, so you can calculate this seperately then use it every time instead of calculating all the time.

chrisbeardy: Also what do you think the ouput of the dictionary would look like for arrays and arrays of structures?

I can imagine something like this for arrays:

```python
{
'bVar': True,
'rVar': 123.456
'iVar': 45
'varName': varValue,
'arrayVar': [1, 2, 3, 4]
}
```

And I guess you would just have a list of dictionaries if you were trying to read an array of dictionaries.

It is also easier to just pass back a list as the way the intended read from the PLC is working is that you just get back a giant list of bytes from the PLC, which is easy to turn into a list, so I think I would have to go through a list of variables anyway when manipulating the bytes into a dictionary?

chrisbeardy: Sorry to bombard you with comments but another potential implememntation I was thinking of would be to do just amend the read_by_name method to do something like:

constants.py
```python
PLCTYPE_STRUCT = PLCTYPE_BYTE
```

ads.py
```python
def read_by_name(name, type, structure_def=None):
    if structure_def is not None:
        do stuff similar to read_structure_by_name implementation
```

users script
```python
structure_def = (
(bla, bla , bla)
)

values = plc.read_by_name('global.samplestruct', pyads.PLCTYPE_STRUCT, structure_def=structure_def)
```
returns
```python
>>>print(values)
[1,2,True, 3.456]
```

Not sure if this would be prefrable to having another standalone method?

stlehmann: As the structure_def in your implementation contains fieldnames it only makes sense to return a dictionary. I don't see why the fieldnames would be needed otherwise. Also I find it more consistent to return a struct as a dict.

Maybe we could use an OrderedDict here. So the user has both possibilities:

    ordered_dict["bVar"] will return the value of a specific field
    list(ordered_dict.values()) will return all values as a list.

What do you think?

chrisbeardy: Yes the field names in my current implementation are not used, I only added them thinking it would be easier to keep track of the structure definition as a user. I did think of an ordered dict but was trying to avoid another import. However I do think it makes sense as a dictionary as you say and this was my first thought.

So use an ordered dict as the output for a read_structure_by_name method? 

If you agree I'll change to this.

stlehmann: I think an OrderedDict would be the right choice here. I don' t  mind the extra import because it is from the standard library so we won't add any new dependencies. So, yes... would be great if you could adjust the implementation to use OrderedDict.

chrisbeardy:
@stlehmann I've updated the implementaion to output a dict. Do you mind taking a look at the overall concept and more specifically the dict_from_bytes function inside ads.py, to see if you have a better idea, as my current code feels a but messy and not very "pythonic", I am not a big fan of the two for loops and the large if/else statement.

Also not sure why it thinks I've relaced the whole files, I'll check the line endings of my editor as I just upgraded pycharm on my linux PC.

On another note, thank you for accepting PRs for the handles and updating pyads version on PyPi, when this new read_structure_by_name functionality is complete, I will add the handle as as input too, I shall then merge your master into my PR so to avoid conflicts,  do you want me just to merge, or would you rather I do a rebase on top of your master then put in another PR instead of this one?

chrisbeardy:
rebased, changed feature to output dictionary so made new PR #92

PR #89, opened 2019-09-16
additional methods for get and release PLC variable handle

chrisbeardy:
I've added in additional methods into the Connection class to get and release PLC variable handles. Additionally I have then modified read and write by name methods and functions to allow the user to pass this handle in as an optional argument.  This gives the user to option to handle getting and releasing of handles themselves and in doing so **speeds the reading and writing process up by up to 3x** when using read_by_name and write_by_name.

As this is an optional argument it should not affect user's existing code or those that do not wish to handle handles themselves, but those that do can get a significant speed advantage.

### DIFF
--- a/pyads/pyads_ex.py
+++ b/pyads/pyads_ex.py
@@ -638,8 +638,42 @@ def adsSyncReadReqEx2(
     return data
 
 
-def adsSyncReadByNameEx(port, address, data_name, data_type, return_ctypes=False):
-    # type: (int, AmsAddr, str, Type, bool) -> Any
+def adsGetHandle(port, address, data_name):
+    # type: (int, AmsAddr, str) -> int
+    """Get the handle of the PLC-variable
+
+    :param int port: local AMS port as returned by adsPortOpenEx()
+    :param pyads.structs.AmsAddr address: local or remote AmsAddr
+    :param data_name: string data_name: data name
+    :return: int: PLC-variable handle
+    """
+    handle = adsSyncReadWriteReqEx2(
+        port,
+        address,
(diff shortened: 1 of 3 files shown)


pyads/pyads_ex.py
```diff
@@ -664,19 +657,57 @@ def adsSyncReadByNameEx(port, address, data_name, data_type, return_ctypes=False
...
+    if no_handle is True:
+        adsReleaseHandle(port, address, handle)
 
     return value
 
 
-def adsSyncWriteByNameEx(port, address, data_name, value, data_type):
-    # type: (int, AmsAddr, str, Any, Type) -> None
+def adsSyncWriteByNameEx(
+    port, address, data_name, value, data_type, handle=None):
```
stlehmann: Please use [black](https://pypi.org/project/black/) for formatting. The closing bracket should be on a separate line like so:

```
def adsSyncWriteByNameEx(
    port, address, data_name, value, data_type, handle=None
):

```

chrisbeardy:
@stlehmann  changes implemented in lastest commit, i'll make sure to run black over my pyads code in the future

stlehmann:
Thanks again for your great work here. I want to ask you if you would also like to add a small chapter in the documentation about using the getHandle and releaseHandle functions. In my opinion a good place for this would be in a chapter Quickstart / Usage / Read and write values by handle. What do you think?

chrisbeardy:
No worries, I have some spare time in between some jobs, and as a Python and Beckhoff user I thought this project was a worthwile use of that time as I've used your library before to good effect, so thanks for all your effort putting that together!  I'll add in a PR for some docs in shortly just before the read and write values by address section.

PR #90, opened 2019-09-18
changed BYTE type to ctypes c_ubyte

### LINKED ISSUE #85: BYTE Type is read with wrong bounds (signed vs unsigned)
When using "read_by_name" the PLC Data Type 'BYTE' should have bounds '0-255', instead values of 128 to 256 come in as -128 to -1.

chrisbeardy:
This fixes #85 , writing was not an issue using c_byte but when reading with c_byte python unpacks this wrong (e.g. 200 reads as -56). The PLC BYTE type is unsigned therefore using c_ubyte is correct.

### DIFF
--- a/pyads/constants.py
+++ b/pyads/constants.py
@@ -13,7 +13,7 @@ from typing import Type
 from ctypes import (
     Array,
     c_bool,
-    c_byte,
+    c_ubyte,
     c_int8,
     c_uint8,
     c_int16,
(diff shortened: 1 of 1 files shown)


stlehmann:
Thanks for the PR. Checked it right away. Nasty mistake. Actually I didn't even know a c_ubyte datatype exists in ctypes

PR #92, opened 2019-09-26
read_structure_by_name

chrisbeardy:
Merged and rebased on top of the master branch, PR replacing PR #88. 

I think this needs review for the overall concept and more specifically the dict_from_bytes function inside ads.py, to see if you have a better idea, as my current code feels a but messy and not very "pythonic", I am not a big fan of the two for loops and the large if/else statement.


chrisbeardy:
I also seem to have broken backward compatibility - shall review

pyads/ads.py
```diff
@@ -418,6 +447,203 @@ def set_timeout(ms):
...
+        for item in structure_def:
+            try:
+                var, plc_datatype, size = item
+                str_len = None
+            except ValueError:
+                var, plc_datatype, size, str_len = item
+
+            if size == 1:  # add straight to dictionary if not array
+                if plc_datatype == PLCTYPE_STRING:
+                    if str_len is not None:
```
stlehmann: Just use if str_len is None and omit the else so there is no double negative condition

pyads/ads.py
```diff
@@ -418,6 +447,203 @@ def set_timeout(ms):
...
+    for structure in range(0, array_size):
+        values = OrderedDict()
+        for item in structure_def:
+            try:
+                var, plc_datatype, size = item
+                str_len = None
+            except ValueError:
+                var, plc_datatype, size, str_len = item
+
+            if size == 1:  # add straight to dictionary if not array
```
stlehmann: Just thinking loudly. I think we could omit the special treatment of arrays by treating array_size=1 also like an array. Just add a check if size==1 at the end and return the first element of the array. Like so:

```python
var_array = []
    for i in range(size):
        # ...
        if plc_datatype in PLC_8_BYTE_TYPES:
            var_array.append(
                struct.unpack(
                    DATATYPE_MAP[plc_datatype],
                    bytearray(byte_list[index : (index + 8)]),
                )[0]
            )
            index += 8
    if size==1:
        values[var] = (var_array[0])
    else:
        values[var] = (var_array)
```

pyads/ads.py
```diff
@@ -418,6 +447,203 @@ def set_timeout(ms):
...
+                        str_len = PLC_DEFAULT_STRING_SIZE
+                    values[var] = (
+                        bytearray(byte_list[index : (index + (str_len + 1))])
+                        .partition(b"\0")[0]
+                        .decode("utf-8")
+                    )
+                    index += str_len + 1
+                elif plc_datatype not in DATATYPE_MAP:
+                    raise RuntimeError("Datatype not found. Check structure definition")
+                elif plc_datatype in PLC_8_BYTE_TYPES:
```
stlehmann: I also think the condition for the different BYTE_TYPES can be put in one block. This
way we get rid of the many if-conditions:

```python
n_bytes = ctypes.sizeof(plc_datatype)
values[var] = struct.unpack(
     DATATYPE_MAP[plc_datatype],
     bytearray(byte_list[index : (index + n_bytes)]),
)[0]
index += n_bytes
```
Both changes should slim down the function a lot. What do you think?

chrisbeardy: Useing ctypes.sizeof is a great move, I knew this could be improved so thank you for that! The changes you have suggested removed a lot of unrequired lines

PR #94, opened 2019-10-02
added CONTRIBUTING.md and updated tox ini to align with travis

chrisbeardy:
I added this in as I think it will help people like me to contribute, mainly to remember to run black!

Feel free to add anymore configurations to ignore for black.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,23 @@
+[tool.black]
+line-length = 88
+target-version = ['py27', 'py35', 'py36']
+include = '\.pyi?$'
+exclude = '''
+
+(
+  /(
+      \.eggs
+    | \.git
+    | \.mypy_cache
+    | \.tox
+    | \.venv
+    | _build
+    | buck-out
+    | build
+    | dist
+    | adslib
+    | pyads.egg-info
+  )/
+  | pyads/__init__.py
+)
+'''
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py27, py35, py36
+envlist = py27, py35, py36, py37
 
 [testenv]
 commands = discover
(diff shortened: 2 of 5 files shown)


chrisbeardy:
it could be worth looking at using [pre-commit](https://pre-commit.com/) too for this.

tox.ini
```diff
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py27, py37
+envlist = py27, py35, py36
```
stlehmann: Including py35 and py36 make sense but why did you remove py37 from the list?

pyproject.toml
```diff
@@ -0,0 +1,23 @@
+[tool.black]
```
stlehmann: What is this .toml file for? I don't know the format ;) Is it just for black formatting?

chrisbeardy: Toml is just another markup language, the pyproject. toml file is a project file that some plugins like black use for configuration, in some cases I've seen it used with setup.py too. I've not used black before, normally used autopep8 but I quite like black! Reading the docs it looks for pyproject.toml for configuration.

tox.ini
chrisbeardy: To align it with the Travis build. As Travis only tests 27, 35 and 36 I thought it makes sense for you tox to use the same? Could add 37 back?

chrisbeardy:
I mainly copy and pasted the contributing from other projects as most are the same really, I also find the deluge bittorrent client a really good repo to steal ideas from....

chrisbeardy:
And when I say steal I mean I really like the way they have there project set up and the tools they use. I think precommit is a really good tool.

tox.ini
stlehmann: Ah OK, I didn' t have in mind that we only test with 3.6 in Travis. I think that was due to Python3.7 not being available in Travis at that time. We should change that. Means adding py37 test to tox and Travis would be the move here, I think.

chrisbeardy: agreed, ill do that

Issue #66, opened 2019-04-04
ADSError: ROUTERERR_PORTALREADYINUSE (1286)


evanmj:
Hi,

Thanks for your work on this.  Do you have an insight on to why I'd be getting error 1286?  My code is as follows:
```
pyads.open_port()
remote_ip = '192.168.1.38'
adr = pyads.AmsAddr('5.59.211.2.1.1',851)
pyads.add_route(adr,remote_ip)
plc = pyads.Connection('5.59.211.2.1.1',851)
plc.open()
```

In my case, 192.168.1.38 is the remote machine, with NetID 5.59.211.2.1.1. 

The 'add_route' method returns "connected to 192.168.1.38", but then plc.open() gives the port in use error.

I noticed the route is not in the routing table of the remote PLC... does it work on remote machines instead of 127.0.0.1.1.1?  I don't see a method for adding the password, so I'm not sure how it would work.  To work around that, like I've done with node-ads in the past, I created a static local route to my machine from the PLC side, but I'm not able to connect, given the error above.

I'm on ubuntu with a win7 TC3 PLC.

Thanks!

evanmj:
Update:

I poked around a bit, but I'm still having trouble.  The latest is "ADSError: Internal Error (1)".

chrisbeardy:
Did you ever get this working, I have just started using and had similar issues, however I have resolved them.

Have you tried running 

`python
import pyads
>>> SENDER_AMS = '1.2.3.4.1.1'
>>> PLC_IP = '192.168.0.100'
>>> USERNAME = 'user'
>>> PASSWORD = 'password'
>>> ROUTE_NAME = 'RouteToMyPC'
>>> HOSTNAME = 'MyPC'
>>> PLC_AMS_ID = '11.22.33.44.1.1'
>>> pyads.add_route_to_plc(SENDER_AMS, HOSTNAME, PLC_IP, USERNAME, PASSWORD, route_name=ROUTE_NAME)
`
from linux to add the route, username and passwork is the PLC.

Then using the follwing to access vars:

'pyhton
import pyads
pyads.open_port()
pyads.set_local_address('add.used.ear.lier')
pyads.close_port()

plc = pyads.Connection('plc_ads', port, 'plc_ip')
plc.open()
'

on linux i get a warning about a port being out of range however it still works. 

you can also use the with statement instead of plc = pyads.Connection.

blindidiotscout:
Hi,
[...]

PR #97, opened 2019-10-08
bug fix for setting up remote route from linux

chrisbeardy:
attempting to resolve issue #66 (or dereiative of) 

The simple solution appears to be to swap hostname and route_name, this works for TWINCAT 3.1.3024. and 3022. However I don't have TWINCAT 2 or 3.1.3020 so not sure how this will affect that. 

@Adrian-at-CrimsonAuzre I'm not sure your thoughts on this? What version of TwinCat did you implement in? 

possibly Beckhoff they have switched these bytes around?

Does anyone know how we would tell what version of TwinCat is running?


chrisbeardy:
thanks for such a quick response @Adrian-at-CrimsonAuzre, I shall maybe at some point have a look at #75 to see if I can get something to work for both TC2 and TC3 but I'm the other way to you, I've not access to TC2 ha

chrisbeardy:
Sorry for my absence, just changed jobs...however that has given me access to TC2 and I can confirm that this fix works for me on TC2, TC3 4022 and 4024.

pyads/ads.py
```diff
@@ -361,15 +361,15 @@ def add_route_to_plc(
     route_name=None,
     added_net_id=None,
 ):
-    # type: (AmsAddr, str, str, str, str, str, AmsAddr) -> None
+    # type: (str, str, str, str, str, str, str) -> bool
```
stlehmann, line 364: That's right, the type hints were wrong, here

tests/test_plc_route.py
```diff
@@ -53,13 +53,13 @@ def plc_route_reciever(self):
 
             data = data[4:]  # Remove protocol bytes
 
-            len_sending_host = struct.unpack("<H", data[:2])[0]  # Length of host name
+            len_route_name = struct.unpack("<H", data[:2])[0]  # Length of PLC password
```
stlehmann, line 56: route name or plc password? There is a contradiction here between the comment and the variable name

stlehmann:
Could you please rebase your PR on the current master branch and add a changelog entry?

chrisbeardy:
I'll take a look at this shortly.

chrisbeardy:
one sec, accident...

chrisbeardy:
sorry, when I squashed, I forgot I had already pushed to a remote, so it caused a merge error, I handled this incorrectly and accidentally pushed all the old commits into my remote. So had to correct this, hence the new PR.

PR #98, opened 2019-10-08
bug fix for setting up remote route from linux

chrisbeardy:
alternative solution from #97  for #66 (and deriative)

This or #97 could be implemented as a quick hotfix for the setting up route from linux issue. 

This one just removes the ability to change the route_name, so is essentially like not using the optional inputs of the function. Breaks backward compaibilty but this is aimed a being a hot_fix.

@Adrian-at-CrimsonAuzre what do you think?

### DIFF
--- a/doc/quickstart.md
+++ b/doc/quickstart.md
@@ -16,6 +16,7 @@ Open a port and create an AmsAddr object for the remote machine.
 >>> import pyads
 >>> pyads.open_port()
 32828
+>>> pyads.set_local_address('1.2.3.4.1.1')
 ```
 Add a route to the remote machine (Linux only - Windows routes must be
 added in the TwinCat Router UI).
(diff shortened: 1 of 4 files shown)


chrisbeardy:
p.s. not sure why 3.7 build failed...may pass if forced built again as odd port is use error

stlehmann:
Did someone so far try one of the two fixes? @chrisbeardy which one would you prefer?

chrisbeardy:
> Did someone so far try one of the two fixes? @chrisbeardy which one would you prefer?

I have't tried although it seems that @blindidiotscout was using TC2 when they got the same error and could fix it by swapping the two still. The simplest fix is #97, just swapping the two around. This may have just been a simple error in the first place.

chrisbeardy:
closing as tested other solution

PR #101, opened 2019-10-21
Added feature to pass QtCore.pyqtSignal to notification decorator. Se…

sondre1988:
Added feature to pass a QtCore.pyqtSignal to the notification decorator for ADS callbacks. Se notifcation decorator usage section in "ads.py" file for simple example on usage.

I have used this in a PyQt5 application where I typically needed a light or value to change based on a variable change inside a Beckhoff CX2040.

### DIFF
--- a/pyads/ads.py
+++ b/pyads/ads.py
@@ -980,8 +980,8 @@ class Connection(object):
         """Set Timeout."""
         if self._port is not None:
             adsSyncSetTimeoutEx(self._port, ms)
-
-    def notification(self, plc_datatype=None):
+    
+    def notification(self, plc_datatype=None, pyqtSignal=None):
         # type: (Optional[Type[Any]]) -> Callable
         """Decorate a callback function.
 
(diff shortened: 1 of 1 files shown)


stlehmann:
Thanks for the PR. I appreciate folks working with this repo and thrying to give something back
by contributing. However I need to take care that the repo does not include too many dependencies and serves a broad variety of use-cases. The use-case of using pyads together with PyQt5 is very obvious. In fact this is the main use-case for me, too. However, I don't want pyads to be specialized to PyQt5. There are many use-cases where no GUI is needed or maybe someone uses another GUI framework like PySide2 or Tkinter. That is why I can't accept your PR because it would bind pyads to PyQt5. Hope you understand.

Regards
Stefan

sondre1988:
Hi,

I get your point :-D But what would you recommend as the simplest way to access data value and use it outside the callback function for ADS notifications in general?

PR #111, opened 2020-01-13
Notification decorator test cases

dxpke:
Implements tests for #68 (and fixes the case where datatype is `None` which crashes currently)


dxpke:
Do you still need Python 2.7 compatibility given it is EOL now?

stlehmann:
Thanks for the PR. Sorry for keeping you waiting so long. We will drop Python 2 support with version 3.2 so your tests should work out fine.

PR #112, opened 2020-02-14
Add getters/setters for connection netid and port

dxpke:
I find it useful to be able to query what netid and port a connection is using. Additionally, I need to be able to create a connection object when I don't know the target netid yet. The connection does not actually use the netid until `open()` is called (the ADS functions themselves are not even limited to interacting with one netid). So this commit also adds the possibility to set/change the netid later on.

### DIFF
--- a/pyads/ads.py
+++ b/pyads/ads.py
@@ -633,6 +633,8 @@ class Connection(object):
     @ams_netid.setter
     def ams_netid(self, netid):
         # type: (str) -> None
+        if self._open:
+            raise AttributeError("Setting netid is not allowed while connection is open.")
         self._adr.netid = netid
 
     @property
@@ -643,6 +645,8 @@ class Connection(object):
     @ams_port.setter
     def ams_port(self, port):
         # type: (int) -> None
+        if self._open:
+            raise AttributeError("Setting port is not allowed while connection is open.")
         self._adr.port = port
 
     def __enter__(self):
--- a/pyads/ads.py
+++ b/pyads/ads.py
@@ -625,6 +625,26 @@ class Connection(object):
         self._open = False
         self._notifications = {}  # type: Dict[int, str]
 
+    @property
+    def ams_netid(self):
+        # type: () -> str
+        return self._adr.netid
+
+    @ams_netid.setter
+    def ams_netid(self, netid):
+        # type: (str) -> None
+        self._adr.netid = netid
+
+    @property
+    def ams_port(self):
+        # type: () -> int
+        return self._adr.port
+
+    @ams_port.setter
+    def ams_port(self, port):
+        # type: (int) -> None
+        self._adr.port = port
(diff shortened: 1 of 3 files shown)


dxpke:
I can't see how the test failures could be related to the commit.

stlehmann:
Please add an entry to changelos. Then we can do the merge :)

pyads/ads.py
```diff
@@ -597,6 +597,26 @@ def __init__(self, ams_net_id, ams_net_port, ip_address=None):
...
 
+    @property
+    def ams_netid(self):
+        # type: () -> str
+        return self._adr.netid
+
+    @ams_netid.setter
+    def ams_netid(self, netid):
+        # type: (str) -> None
+        self._adr.netid = netid
```
stlehmann, line 608: It shouln't be possible to change the own net id or port after a connection is opened. Otherwise this is not consistent. Could you add a check here if a conneciton has been opened?

pyads/ads.py
```diff
@@ -597,6 +597,26 @@ def __init__(self, ams_net_id, ams_net_port, ip_address=None):
...
+
+    @property
+    def ams_port(self):
+        # type: () -> int
+        return self._adr.port
+
+    @ams_port.setter
```
stlehmann, line 615: The same as above

tests/test_connection_class.py
```diff
@@ -92,6 +92,16 @@ def test_open_twice(self):
...
+    def test_netid_port(self):
+        self.assertEqual(self.plc.ams_netid, TEST_SERVER_AMS_NET_ID)
+        self.assertEqual(self.plc.ams_port, TEST_SERVER_AMS_PORT)
+        with self.assertRaises(ValueError):
+            self.plc.ams_netid = "1.1.1.1.1.1.1"
+        self.plc.ams_netid = "1.1.1.1.1.1"
+        self.assertEqual(self.plc.ams_netid, "1.1.1.1.1.1")
+        self.plc.ams_port = 1
+        self.assertEqual(self.plc.ams_port, 1)
+
```
stlehmann, line 104: Please add a testcase where you write net-id and port of a connection that is already open. An attribute error should be raised.

dxpke:
@stlehmann as a general FYI I am a bit preoccupied with a customer project at the moment, and so it may take until end of May until I can thoroughly look at all open PRs again. But I will definitely do it.

stlehmann:
No worries. I was able to merge your other PRs so if I can find the time I can have a closer look on this one as well.

dxpke:
Thanks for finishing and merging. I originally did not add the check because on the protocol level it does not matter if you change the target netid/port even when the connection is "open" - the connection does not have a defined endpoint. But from the user POV it makes sense to disallow this because it might cause problems when adding/removing device notifications, for example.

PR #113, opened 2020-02-14
read/readwrite: Add option not to check data size

dxpke:
Some structures (such as SAdsSymbolEntry) are dynamically sized, so the
length cannot be checked with the read data size. Allow the user to
disable length checking on request.


stlehmann:
Almost there. Could you please add an entry to the changelog that describes the changes. Then we are ready to merge.

PR #114, opened 2020-02-14
Allow to write ctypes values directly

dxpke:
If values to be written are already given as ctypes values, allow to
write them directly. This is useful, for example, for writing ctypes
structs and struct arrays.


dxpke:
Test added

stlehmann:
Thanks for this PR. Looks good to me. Sorry for the late answer.

PR #118, opened 2020-03-05
Add support for arrays in notification decorator

dxpke:
Incorporates/depends on #111 for test cases, aa0d8c1 is the only relevant commit here


stlehmann:
CI fails for Python 2.7. Could you please check if there's a quick fix for that. We will omit Python 2.7 support in the next releases anyway. But maybe it's just some tiny bit of code that needs to be changed.

stlehmann:
Thanks for the PR 👍

PR #120, opened 2020-04-09
Add LICENSE and README.rst to the package manifest

klauer:
For packaging pyads on conda-forge, it would be a good idea to include both these files. Then you won't have to include LICENSE inside of the recipe (https://github.com/conda-forge/staged-recipes/pull/10809).


stlehmann:
Thanks, I agree with that. Thanks also for your support on conda-forge. Hopefully this will bring the staging-process forward.

PR #130, opened 2020-05-04
Allow read_write with NULL read/write data

dxpke:
This is useful e.g. for a generic RPC interface where some functions
to be called over ADS do not have any parameter or return value (`void(void)` in C terms).

This is the last PR that I still had in the pipeline :-)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## Unreleased
+
+### Added
+* [#130](https://github.com/stlehmann/pyads/pull/130) Allow read_write with NULL read/write data
+
 ## 3.2.0
 
 ### Added
--- a/pyads/ads.py
+++ b/pyads/ads.py
@@ -778,19 +778,18 @@ class Connection(object):
         return_ctypes=False,
         check_length=True,
     ):
-        # type: (int, int, Type, Any, Type, bool, bool) -> Any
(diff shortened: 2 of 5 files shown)


dxpke:
I can add the ChangeLog entry of course, but I have to note that the workflow is quite awkward. I first have to create the PR so I get the PR number. Only then can I update the Changelog. And need to push an additional commit to update it. Also, from that point on, creating more changelog entries may result in merge conflicts. In my opinion, this is something that should be handled automatically in some fashion (or by the maintainer).

stlehmann:
I don't see a problem in adding another commit to an existing PR. In fact this is a common practice. After all there most certlainly will be changes after the code has been reviewed.

dxpke:
I have no problem with adding commits in general. I just disagree that it should be part of the expected workflow when a PR could otherwise be merged immediately.

dxpke:
Anyway, your project - your rules. Just giving my opinion :-)

stlehmann:
In my opinion updating the Changelog should be done within the PR. This way it is always guaranteed that a PR gets an entry in the Changelog and none gets forgotten.

stlehmann:
But, yeah it is free to you to disagree, here ;)

dxpke:
@stlehmann Something else altogether - I think you may have mistagged the 3.2.0 release as 3.1.1.1

stlehmann:
> Something else altogether - I think you may have mistagged the 3.2.0 release as 3.1.1.1

Thanks for the hint. I mistagged 3.2.0 earlier and thought I fixed it right. Clearly not. Now it should be alright.

dxpke:
@stlehmann updated

PR #137, opened 2020-05-10
Update adslib from upstream

stlehmann:


stlehmann:
Looks like all is well. Merging in master.

PR #138, opened 2020-05-11
Normalize the line endings

tboegi:
State that all files should have LF line endings in the repo.
In the working tree they will have CRLF under Windows and LF under
Linux or MacOs.
This avoids future confusion when file change line endings "on disk"
because the editor uses line fieed when saveing files.


tboegi:
@stlehmann: Merging this patch would show me as the author of nearly all lines -
unless you run "git blame -w"
Compare e.g
git blame  PKG-INFO
vs
git blame -w PKG-INFO

So feel free to merge this PR "as is" to to re-do the normalizing under your name.

stlehmann:
Thanks for the hint and the PR. How did you even find out there was an issue the line-endings. I develop on Mac OS and Windows and didn't even recognize this.

stlehmann:
Thanks again for this PR. As you suggested I renormalized the line-endings under my name to avoid confusion.

tboegi:
The hint was given by emacs saying "DOS" under Linux.
The more official hint can be digged into by using
`git ls-files --eol`

Issue #143, opened 2020-07-06
Notification parsing


corylutton:
Thanks for such a useful library.  I am in the middle of using it to subscribe to tags and forward the messages to MQTT.  In doing so, I noticed that converting the notification to the python object is available via a decorator.  Since I keep a subscriptions dict with their datatype, I have copied the guts of that notification decorator in order to do conversion as needed which simplifies the amount of code required.. my question is if you would be interested in a pull request to have the notification parsing available without using the decorator syntax?

For example you show using the decorator like this:

```
@plc.notification(pyads.PLCTYPE_STRING)
def callback(handle, name, timestamp, value):
    print(handle, name, timestamp, value)
```

I'm proposing making an adjustment that would allow use like this as well:

```
def callback(notification, data):
    datatype = my_subscriptions(data)
    result = self.notification_parse(notification, datatype, data)
    # result here would be (contents.hNotification, data_name, timestamp, value) which would preserve current funtion
```

I would probably just move the guts of the decorator to a helper function called notification_parse on the connection instance which would just open that bit of code for decorator or the use case I propose.  This might also make it easier in the future to automatically do that conversion by keeping track of the notifications requested...

Please let me know if there would be interest in a pull request like that and thanks again for the useful library.

Thanks
Cory Lutton

stlehmann:
@corylutton I'm glad you like the library. I would gladly accept a PR concerning this matter so feel free to go for it.

PR #144, opened 2020-07-08
Add notification_parser method to Connection

corylutton:
As mentioned in #143 - here is my take on moving the logic from the decorator into a different method to be used in other ways + in the decorator.

Hopefully this is up to par, I tried to get the documentation/type info in there as the rest of the code had it... not really that familiar with this process (Github) so could always take another run at it if needed.  I left all the tests alone as all the notification decorator tests just call this function, was not sure if I should poke into that at all or what would make sense to change that into since the decorator now calls this function.

### DIFF
--- a/pyads/ads.py
+++ b/pyads/ads.py
@@ -1149,7 +1149,7 @@ class Connection(object):
         >>> # Create callback function that prints the value
         >>> def mycallback(notification, data):
         >>>     data_type = tag[data]
-        >>>     handle, timestamp, value = plc.notification_parser(notification, data_type)
+        >>>     handle, timestamp, value = plc.parse_notification(notification, data_type)
         >>>     print(value)
         >>>
         >>> with plc:
--- a/pyads/ads.py
+++ b/pyads/ads.py
@@ -633,7 +633,9 @@ class Connection(object):
     def ams_netid(self, netid):
         # type: (str) -> None
         if self._open:
-            raise AttributeError("Setting netid is not allowed while connection is open.")
+            raise AttributeError(
+                "Setting netid is not allowed while connection is open."
+            )
         self._adr.netid = netid
 
     @property
@@ -645,7 +647,9 @@ class Connection(object):
     def ams_port(self, port):
         # type: (int) -> None
         if self._open:
-            raise AttributeError("Setting port is not allowed while connection is open.")
+            raise AttributeError(
+                "Setting port is not allowed while connection is open."
+            )
         self._adr.port = port
 
     def __enter__(self):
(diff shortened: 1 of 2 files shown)


pyads/ads.py
```diff
@@ -633,7 +633,9 @@ def ams_netid(self):
     def ams_netid(self, netid):
         # type: (str) -> None
         if self._open:
-            raise AttributeError("Setting netid is not allowed while connection is open.")
```
stlehmann, line 636: No need to change that. It is not related to the issue.

pyads/ads.py
```diff
@@ -645,7 +647,9 @@ def ams_port(self):
     def ams_port(self, port):
         # type: (int) -> None
         if self._open:
-            raise AttributeError("Setting port is not allowed while connection is open.")
+            raise AttributeError(
```
stlehmann, line 650: Same as above

pyads/ads.py
```diff
@@ -1107,44 +1111,92 @@ def notification_decorator(func):
...
-                    memmove(addressof(value), addressof(data), fit_size)
+        return notification_decorator
 
-                elif plc_datatype is not None and issubclass(plc_datatype, Array):
-                    if data_size == sizeof(plc_datatype):
-                        value = list(plc_datatype.from_buffer_copy(bytes(data)))
-                    else:
-                        # invalid size
-                        value = None
+    def notification_parser(
```
stlehmann, line 1123: I would prefer the function name to be a verb because it is no object but does something. Maybe something like "parse_notification"

corylutton: Makes sense, parse_notification sounds good to me

corylutton:
Addressed your comments above, I think I got it now without additional unrelated changes.  Not sure if there is a way to combine those three commits into one or if it matters.  If I can do anything else to make it better please let me know... long time programmer but only 3rd time doing a github pull request.

pyads/ads.py
```diff
@@ -1107,44 +1107,92 @@ def notification_decorator(func):
...
+        >>> from ctypes import size_of
+        >>>
+        >>> # Connect to the local TwinCAT PLC
+        >>> plc = pyads.Connection('127.0.0.1.1.1', 851)
+        >>> tag = {"GVL.myvalue": pyads.PLCTYPE_INT}
+        >>>
+        >>> # Create callback function that prints the value
+        >>> def mycallback(notification, data):
+        >>>     data_type = tag[data]
+        >>>     handle, timestamp, value = plc.notification_parser(notification, data_type)
```
stlehmann, line 1152: I suppose this should be plc.parse_notification
```suggestion
        >>>     handle, timestamp, value = plc.parse_notification(notification, data_type)
```

corylutton: Documentation adjusted.

stlehmann:
Great work. Thanks for this 🎉

Issue #124, opened 2020-04-23
insert AdsReadSymbolInfo


pigghino:
hi,
can you insert AdsReadSymbolInfo in this library?

Thanks

stlehmann:
I flagged this as enhancement. You are very welcome to contribute to this package. Maybe you want to make a PR?

stlehmann:
This issue is adressed by https://github.com/stlehmann/pyads/pull/155 so I will close it.

PR #155, opened 2020-09-19
Add get_all_symbols function

chrisjbremner:
I was looking for a way to retrieve all symbol information and came across https://github.com/stlehmann/pyads/issues/124, indicating this was not a feature. I searched the ADS repo for `AdsReadSymbolInfo` which https://github.com/stlehmann/pyads/issues/124 implies exists, but I could not find it.

The ability to read symbols was a feature I had used before in the [`counsyl-pyads`](https://github.com/counsyl/counsyl-pyads) package, so I decided to port over [the logic](https://github.com/counsyl/counsyl-pyads/blob/e6fafd946b894c22e555a09c97fcd301844f212a/counsyl_pyads/adsclient.py#L313-L358). There's also logic in that package that allows you to read a symbol by name, but I decided against adding that for now unless there's a request to do so.

Here's a sample when tested with a PLC running TC3:
```
>>> symbols=plc.get_all_symbols()
>>> print('\n'.join("%s: %s" % item for item in vars(symbols[0]).items()))
index_group: 16448
index_offset: 384800
name: Constants.bFPUSupport
symtype: BOOL
comment: Does the target support multiple cores?
```

You may also notice that the import ordering was slightly modified to be in alphabetical order by `isort`. If this is not a desired change, I can revert it.

### DIFF
--- a/pyads/__init__.py
+++ b/pyads/__init__.py
@@ -32,6 +32,7 @@ from .constants import (
     PLCTYPE_DT,
     PLCTYPE_DWORD,
     PLCTYPE_INT,
+    PLCTYPE_LINT,
     PLCTYPE_LREAL,
     PLCTYPE_REAL,
     PLCTYPE_SINT,
--- a/pyads/utils.py
+++ b/pyads/utils.py
@@ -12,8 +12,7 @@
 import functools
 import sys
 import warnings
-
-from typing import Callable, Any, Optional
+from typing import Any, Callable, Optional
 
 
 def platform_is_linux() -> bool:
@@ -46,3 +45,8 @@ def deprecated(message: Optional[str] = None) -> Callable:
         return wrapper
 
     return decorator
+
+
+def decode_ads(message: bytes) -> str:
+    """Decode a string that in encoded in the format used by ADS."""
+    return message.decode("windows-1252").strip(" \t\n\r\0")
--- a/pyads/utils.py
+++ b/pyads/utils.py
@@ -49,5 +49,12 @@ def deprecated(message: Optional[str] = None) -> Callable:
 
 
 def decode_ads(message: bytes) -> str:
-    """Decode a string that in encoded in the format used by ADS."""
+    """
+    Decode a string that in encoded in the format used by ADS.
+
+    From Beckhoff documentation: 'A STRING constant is a string enclosed by
+    single quotation marks. The characters are encoded according to the Windows
+    1252 character set. As a subset of Windows-1252, the character set of
+    ISO/IEC 8859-1 is supported.'
+    """
     return message.decode("windows-1252").strip(" \t\n\r\0")
--- a/pyads/utils.py
+++ b/pyads/utils.py
@@ -12,7 +12,8 @@
 import functools
 import sys
 import warnings
-from typing import Any, Callable, Optional
+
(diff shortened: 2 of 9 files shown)


pyads/utils.py
```diff
@@ -12,8 +12,7 @@
 import functools
 import sys
 import warnings
-
-from typing import Callable, Any, Optional
+from typing import Any, Callable, Optional
```
stlehmann, line 15: The same as above

stlehmann (changes requested):
Thanks for the PR. I find this to be a useful feature. So far it looks good for me. Please have a look at my review for minor change requests. Of course there also need to be some tests and documentation for the new feature. Would you like to take care of these, too?

pyads/__init__.py
```diff
@@ -32,6 +32,7 @@
     PLCTYPE_DT,
     PLCTYPE_DWORD,
     PLCTYPE_INT,
+    PLCTYPE_LINT,
```
chrisjbremner, line 35: I noticed that this was missing from the imports. Not strictly needed for this PR, so I can remove it from here if you'd prefer it in a separate PR.

chrisjbremner:
Thanks for taking a look, I've addressed your comments and added some tests and some documentation to the quickstart guide.

Regarding the import re-ordering, let me know if you'd prefer me to open a separate PR for reordering imports or if you'd like to leave it as is.

pyads/utils.py
```diff
@@ -46,3 +46,8 @@ def wrapper(*args: Any, **kwargs: Any) -> Callable:
         return wrapper
 
     return decorator
+
+
+def decode_ads(message: bytes) -> str:
+    """Decode a string that in encoded in the format used by ADS."""
+    return message.decode("windows-1252").strip(" \t\n\r\0")
```
stlehmann, line 53: Is this the general encoding Beckhoff uses for strings?

stlehmann (changes requested):
Looks great. Would you mind adding an entry to the Changelog. Then the PR is ready to merge.

pyads/utils.py
chrisjbremner: I believe so, according to [this](https://infosys.beckhoff.com/english.php?content=../content/1033/tc3_plc_intro/9007201784068235.html&id=):

> A STRING constant is a string enclosed by single quotation marks. The characters are encoded according to the Windows 1252 character set. As a subset of Windows-1252, the character set of ISO/IEC 8859-1 is supported.

stlehmann (approved):

pyads/__init__.py
stlehmann: Never mind. This is just a small import. No need to make a separate PR for it.

chrisjbremner:
Thanks for taking a look. I'm not able to merge, so feel free to do so yourself when you get a chance.

stlehmann:
Thanks @chrisjbremner for the great work 👍

chrisbeardy:
this is a great contribution!!!!

chrisjbremner:
@stlehmann could I request a new release that incorporates this addition?

stlehmann:
@chrisjbremner Yes, I think now is a good time for a new major release. I will get to it.

stlehmann:
Release 3.3.0 is out.

PR #157, opened 2020-10-02
Added write_by_dict functionality

chrisbeardy:
- includes function to convert to bytes from an OrderedDict
- includes tests and docs

@stlehmann not sure if you want to tidy up or rethink the bytes_from_dict function, to make it look nicer, less complex? Can you think of any improvements to it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 * [#155](https://github.com/stlehmann/pyads/pull/155) Add get_all_symbols method to Connection
+* [#157](https://github.com/stlehmann/pyads/pull/157) Add write_structure_by_name method to Connection
 
 ### Changed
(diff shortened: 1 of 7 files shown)


chrisbeardy:
@stlehmann not sure why the tests are failing now, they seem to be failing on `test_get_all_symbols_empty` which I have not changed and seems to be an issue with the ADS server? Are my new tests impacting this test?

chrisjbremner:
That was a test I introduced in https://github.com/stlehmann/pyads/pull/155. I suspect that adding a ``time.sleep(1)` after starting the server in `setUpClass` like the `AdsConnectionClassTestCase` does would fix it. Let me give it a try in a new PR.

stlehmann (changes requested):
@chrisbeardy This looks great to me. Would you mind changing the base branch for the merge to "dev"? And if you like you can add an entry to the changelog that describes the new feature.

stlehmann (approved):
Looks great now. Thanks for your engagement 👍

Issue #158, opened 2020-10-02
Struggle to run tests on Windows laptop


chrisbeardy:
I struggle to run the ADS test server on my windows laptop, I often get the follwing:

```
  File "C:\Users\chrisb\GitWS\pyads\tests\test_connection_class.py", line 39, in setUpClass
    cls.test_server = AdsTestServer(logging=True)
  File "c:\users\chrisb\gitws\pyads\.tox\py37\lib\site-packages\pyads\testserver.py", line 108, in __init__
    self.server.bind((self.ip_address, self.port))
OSError: [WinError 10013] An attempt was made to access a socket in a way forbidden by its access permissions
```

I also get the following for Python3.8

```
ERROR: test_ads (discover.ModuleImportFailure)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "c:\users\chrisb\gitws\pyads\.tox\py38\lib\site-packages\discover.py", line 64, in testFailure
    raise exception
ImportError: Failed to import test module: test_ads
Traceback (most recent call last):
  File "c:\users\chrisb\gitws\pyads\.tox\py38\lib\site-packages\discover.py", line 288, in _find_tests
    module = self._get_module_from_name(name)
  File "c:\users\chrisb\gitws\pyads\.tox\py38\lib\site-packages\discover.py", line 266, in _get_module_from_name
    __import__(name)
  File "C:\Users\chrisb\GitWS\pyads\tests\test_ads.py", line 9, in <module>
    import pyads
  File "c:\users\chrisb\gitws\pyads\.tox\py38\lib\site-packages\pyads\__init__.py", line 11, in <module>
    from .ads import (
  File "c:\users\chrisb\gitws\pyads\.tox\py38\lib\site-packages\pyads\ads.py", line 19, in <module>
    from .pyads_ex import (
  File "c:\users\chrisb\gitws\pyads\.tox\py38\lib\site-packages\pyads\pyads_ex.py", line 65, in <module>
    _adsDLL = ctypes.WinDLL("TcAdsDll.dll")  # type: ignore
  File "C:\Users\chrisb\AppData\Local\Programs\Python\Python38\lib\ctypes\__init__.py", line 373, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: [WinError 193] %1 is not a valid Win32 application
```

Does someone have any insight to why this happens?

chrisjbremner:
[...]

PR #159, opened 2020-10-02
General tdy up of tests and changes to reflect depreication of old functions

chrisbeardy:
I have created this PR as a general tidy up, i have seperated out the commits so hopefully each commit shows what it does, I can seperate into different PRs if required but didn't want to create too many.

Changes:
- fixed a test, although it was passing, I realised the bytes values were wrong after the partition value.
- changes to contributing and pyproject.toml to reflect depriciation of functions and python2
- modified some of the structure examples as they were quite long


chrisbeardy:
failed only only the `get_all_symbols_empty`, I don't think there is an issue with that test but probably just the ADS server on the build system?

chrisbeardy:
I have added reference in Contributing to using Docker for tests on windows should there be an issue with the ADS Test Server as in #158  .

I have also amended the Dockerfile so images can easily be created for multiple python versions to cover all tox test cases. I've also switched to using pytest instead of `setup.py test` as the latter is being depreciated. I hope this is useful.

stlehmann (approved):
Thanks for tidying up 👍

PR #160, opened 2020-10-02
Add a sleep after starting advanced handler server

chrisjbremner:
Adds a sleep to address the test error @chrisbeardy noted in https://github.com/stlehmann/pyads/pull/157 and https://github.com/stlehmann/pyads/pull/159


stlehmann (approved):
@chrisjbremner Thanks for this quick fix. I was already wondering why the tests were failing

PR #161, opened 2020-10-25
Add sum commands (2nd PR)

bdcoyle:
This PR supersedes #149 
It provides functions 'read_list_by_name' and 'write_list_by_name', which allow a list of variables to be read or written in a single call.
The functions get the symbol info from the ADS server, so type is not required to be provided. The symbol info can be cached to make subsequent read/writes more efficient.


pyads/pyads_ex.py
```diff
@@ -736,6 +780,165 @@ def adsGetHandle(port: int, address: AmsAddr, data_name: str) -> int:
...
+                data_symbols[data_name].dataType != ADST_STRING
+                and data_symbols[data_name].dataType != ADST_WSTRING
+            ):
+                value = struct.unpack_from(
+                    DATATYPE_MAP[ads_type_to_ctype[data_symbols[data_name].dataType]],
+                    sum_response,
+                    offset=data_start + offset,
+                )[0]
+            else:
+                print("DATA_NAME: ", data_name)
```
stlehmann, line 860: We should not use print to avoid verbose outputs. Was this for debugging during development only then you may want to remove it.

stlehmann (approved):

stlehmann:
Looks much better now 👍. I found only some minor issues. Then this PR is ready for merging.

pyads/ads.py
```diff
@@ -371,6 +377,7 @@ def __init__(
             self.ip_address = ip_address
         self._open = False
         self._notifications = {}  # type: Dict[int, str]
+        self._symbol_info_cache = {} # type: Dict[str, SAdsSymbolEntry]
```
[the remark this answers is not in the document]
bdcoyle: Thanks, fixed

pyads/pyads_ex.py
bdcoyle: That's right, was only for development and I missed it when tidying up. Thanks

stlehmann:
Thanks a load again for your effort with the sum-commands. I think this is a huge improvement and look forward on testing them myself 👍 😁

chrisbeardy:
I agree, this is awesome, it's going to make things a lot easier.

pylipp:
Thank you a lot for this extension!

I wonder whether this is compatible with structured ADS variables? E.g. 
```
>>> structure_def = (
...    ('rVar', pyads.PLCTYPE_LREAL, 1),
...    ('rVar2', pyads.PLCTYPE_REAL, 1),
...    ('iVar', pyads.PLCTYPE_INT, 1),
...    ('iVar2', pyads.PLCTYPE_DINT, 3),
...    ('sVar', pyads.PLCTYPE_STRING, 1)
... )
>>> plc.read_list_by_name(['global.sample_structure'])
```

bdcoyle:
Currently it is only supporting native ADS types. I can take a look into it to see what would be needed for it to support structs

PR #169, opened 2020-11-19
Add method to get AMS Net ID of PLC from IP

chrisjbremner:
I find it a bit unintuitive to have to provide the AMS Net ID of the PLC at connection time in this library, whereas when you use the TwinCAT ADS router, all you need to provide is an IP address, and it finds the AMS Net ID for you. To get around this, I took a similar approach to https://github.com/stlehmann/pyads/pull/75, and sniffed the UDP request that the ADS Router makes when it discovers the AMS Net ID from a PLC, and wrote a function to execute that request for me. Here's what it looks like in use:

```python
>> getNetIdForPLC("192.168.0.3")

'1.2.3.4.5.6'
```

I can imagine this funciton being useful to call inside of `add_route_to_plc` and the constructor for `Connection` if `ams_net_id`/`ams_net_port` are not provided.

The request is pretty similar to the one made in `adsAddRouteToPLC`, so I was able to re-use most of that logic. The request also returns additional discovery information that's available, such as the version of TwinCAT running on the PLC, the OS version, and the fingerprint -- at the moment I make no use of this information.

I created this PR in an unfinished state to get feedback on whether this is a useful feature, and whether this approach makes sense. If there's consensus that I should continue, I will continue working on it (and improving code re-use) and then mark the PR as ready for review at that time.

Note that I tested this function on a Windows machine and haven't tested it yet in Linux, but I believe it should work for both operating systems.


chrisjbremner:
@stlehmann I'd be curious to get your thoughts on this proposal

stlehmann:
@chrisjbremner sorry for keeping you waiting so long :camel:

I definetely like your idea of simplifying the add-route procedure. I can imagine a modified `add_route()` function  that
calls `getNetIdForPlc()` if `adr` is None.

So, yes ... I would love to see you elaborate on that PR. :+1:

chrisjbremner:
Ok, I went ahead and added this feature to `ads.py`. Maintaining compatibility with previous versions was a bit difficult, so I didn't re-arrange any of the args/kwargs and I made do with them the best I could

pyads/ads.py
```diff
@@ -178,10 +179,13 @@ def set_local_address(ams_netid: Union[str, SAmsNetId]) -> None:
 def add_route(adr: Union[str, AmsAddr], ip_address: str) -> None:
     """Establish a new route in the AMS Router (linux Only).
 
-    :param adr: AMS Address of routing endpoint as str or AmsAddr object
+    :param adr: AMS Address of routing endpoint as str or AmsAddr object. If an
```
stlehmann, line 182: I propose we use None to indicate that no ams address is provided. So the data type should be Optional[Union[str, AmsAddr]]

pyads/ads.py
```diff
@@ -178,10 +179,13 @@ def set_local_address(ams_netid: Union[str, SAmsNetId]) -> None:
...
+        empty string is provided, the net id of the PLC will be discovered.
     :param str ip_address: ip address of the routing endpoint
 
     """
+    if not adr:
```
stlehmann, line 187: Check for None here

pyads/pyads_ex.py
```diff
@@ -176,6 +177,33 @@ def adsAddRoute(net_id: SAmsNetId, ip_address: str) -> None:
...
+) -> Tuple[bytes, str]:
+    """Send a raw UDP message to the PLC and return the response
+
+    :param str ip_address: ip address of the PLC
+    :param message: the message to send to the PLC
+    :expected_return_length: number of bytes to expect in response
+    """
+    with closing(socket.socket(socket.AF_INET, socket.SOCK_DGRAM)) as sock:  # UDP
+        # Listen on 55189 for the response from the PLC
+        sock.bind(("", 55189))
```
stlehmann, line 191: Just out of curiosity: Is every TwinCAT plc sending data from 55189 and receiving on 48899?

stlehmann: How about error handling? What happens if you can not bind to that port?

stlehmann (changes requested):
Looks greate. I' ve just got some small change-requests and questions.
Great work so far.

pyads/pyads_ex.py
chrisjbremner: For both of these, I just copied from #75. Regarding port 55189 -- the sending port doesn't actually matter, but the receiving port does, because that's where [ADS expects UDP data to be sent to](https://infosys.beckhoff.com/content/1033/tcremoteaccess/html/tcremoteaccess_firewall.html?id=4860001493050562785). I removed the port 55189 requirement and instead changed it to [bind to any available port using `0`](https://stackoverflow.com/a/1365284/4409369), and I updated the tests to reflect this as well.

pyads/ads.py
chrisjbremner: Sounds good, unfortunately we cannot give it a default value because of the argument ordering (so as to maintain backwards compatibility), but we can at least change the type annotation.

pyads/pyads_ex.py
```diff
@@ -176,6 +177,34 @@ def adsAddRoute(net_id: SAmsNetId, ip_address: str) -> None:
...
+    :param message: the message to send to the PLC
+    :expected_return_length: number of bytes to expect in response
+    """
+    with closing(socket.socket(socket.AF_INET, socket.SOCK_DGRAM)) as sock:  # UDP
+        # Listen on any available port for the response from the PLC
+        sock.bind(("", 0))
+
+        # Send our data to 48899 on the PLC, which is where ADS expects UDP
+        # messages
+        sock.sendto(message, (ip_address, 48899))
```
stlehmann, line 195: We should not have hard-coded ports in the code. I think it would be nice to put this port number into [constants.py](https://github.com/stlehmann/pyads/blob/440d0a18493000c407cc2e65a1c795902948ac36/pyads/constants.py#L189) in the PORTS section. Something like PORT_REMOTE_UDP.

tests/test_plc_ams.py
```diff
@@ -0,0 +1,49 @@
...
+
+class PLCAMSTestCase(unittest.TestCase):
+
+    PLC_IP = "127.0.0.1"
+    PLC_AMS_ID = "11.22.33.44.1.1"
+
+    def plc_ams_request_receiver(self):
+        with closing(socket.socket(socket.AF_INET, socket.SOCK_DGRAM)) as sock:
+            # Listen on 48899 for communication
+            sock.bind(("", 48899))
```
stlehmann, line 18: Please use the global port variable here.

stlehmann (changes requested):

pyads/pyads_ex.py
```diff
@@ -176,6 +178,33 @@ def adsAddRoute(net_id: SAmsNetId, ip_address: str) -> None:
         raise ADSError(error_code)
 
 
+def send_raw_udp_message(
+    ip_address: str, message: bytes, expected_return_length: int
+) -> Tuple[bytes, str]:
```
stlehmann, line 183: Please add :rtype: and :returns: to docstring

pyads/pyads_ex.py
```diff
@@ -195,12 +224,9 @@ def adsAddRouteToPLC(
     :param str password: password for PLC
     :param str route_name: PLC side name for route, defaults to adding_host_name or the current hostname of this PC
     :param pyads.structs.SAmsNetId added_net_id: net id that is being added to the PLC, defaults to sending_net_id
+    :returns: True if the provided credentials are correct, False otherwise
```
stlehmann, line 227:
```suggestion
     :rtype: bool
     :returns: True if the provided credentials are correct, False otherwise
```

pyads/pyads_ex.py
```diff
@@ -285,6 +299,36 @@ def adsAddRouteToPLC(
     raise RuntimeError("Unexpected response from PLC")
 
 
+def adsGetNetIdForPLC(ip_address: str) -> Optional[str]:
+    """Get AMS Net ID from IP address.
+    
+    :param str ip_address: ip address of the PLC
+    :returns: net id of the device at the provided ip address
```
stlehmann, line 306:
```suggestion
    :rtype: Optional[str]
    :returns: net id of the device at the provided ip address
```

pyads/pyads_ex.py
```diff
@@ -285,6 +299,36 @@ def adsAddRouteToPLC(
     raise RuntimeError("Unexpected response from PLC")
 
 
+def adsGetNetIdForPLC(ip_address: str) -> Optional[str]:
```
stlehmann, line 302: Is there a way this function returns None? I see it like this: It either returns the net id or raises an Exception. So the return type should be non-optional

stlehmann (changes requested):

stlehmann (changes requested):
Looks great so far. Now all we need is an entry in the Changelog and you are ready to merge.

stlehmann:
Thanks @chrisjbremner for this nice PR.

PR #172, opened 2020-12-06
Apply black to files that were missing it

chrisjbremner:
I noticed that `black` was catching a few more fixes in this codebase, so I decided to push them in their own PR.


stlehmann (approved):

stlehmann:
Thanks for this PR 👍

Issue #168, opened 2020-11-17
Problems with getting symbol info from AdsTestServer


RobertoRoos:
I am writing the tests for #166 and having some trouble with getting it to work with the test server.

I copied the `AdsConnectionClassTestCase` test class and named it `AdsSymbolTestCase`.

Test code like this:

```
        with self.plc:
            symbol = AdsSymbol(self.plc, handle_name)
            print('Symbol:', symbol)
            print('Value:', symbol.value)
```

....fails because the type could not be determined using `pyads_ex.adsGetSymbolInfo`. The `type_name` field of the info struct is an empty string.

How can we populate the `AdsTestServer` with test variables?  
How does `AdsApiTestCaseAdvanced.test_get_all_symbols_single` create the variable that is being returned?

Also in `AdsConnectionClassTestCase.test_read_by_name` a variable with the name "TestHandle" is read from. Where is this variable created?

stlehmann:
@RobertoRoos is this still an issue? I just came across this issue but couldn't find the tests you were referring to.

RobertoRoos:
Yeah, haven't made much progress.

I see haven't pushed any tests yet, but haven't written anything useful yet.

It think it's best to ask it as an open question: How can I make `adsGetSymbolInfo()` return a PLCTYPE from the testserver?
The `BasicHandler` always returns 0, which does not correspond to an actual.

RobertoRoos:
Making some progress. Basically, there was no support for a `ADSIGRP_SYM_INFOBYNAMEEX` type request in the `AdvancedHandler`. Currently writing this.

And to answer my original question: `AdvancedHandler` remembers the variables addressed in previous requests. So you can create remote variables simply writing something to it.

stlehmann:
@RobertoRoos Just to clean it up: Has this issue been resolved with your latest PR?

RobertoRoos:
Yes, the current `AdsSymbolTestCase` gives an example of how this works.

PR #174, opened 2020-12-11
Add AdsSymbol class

### LINKED ISSUE #166: ADS Symbol class
Let me start that I am loving this library! This is allowing my team and me to easily create an application around a TwinCAT solution in good 'ol familiar Python.

For a part of our application we created a Symbols class that wraps around an ADS variable. The goal was to reduce:

```python
index_group, index_offset = ...
plc.read(index_group, index_offset, pyads.PLCTYPE_LREAL)
plc.write(index_group, index_offset, 420.0, pyads.PLCTYPE_LREAL)
```

...to:

```python
symbol = Symbol(plc, ...)
symbol.get()
symbol.set(420.0)
```

The Symbol class keeps a reference to the Connection and remembers the indices. (The example is a little trivial, but for the large number of variables in a tree structure this became very helpful.)

Would it be worthwhile to make such a class part of pyads itself?

RobertoRoos:
Expanded the `AdsSymbol` class to stand on its own. Also updated the `AdvancedHandler` tester class.
Fixes #166 and related to #168


pyads/ads.py
```diff
@@ -671,6 +682,11 @@ def read(
 
         return None
 
+    def get_symbol(self, name: str) -> AdsSymbol:
+        """Get a single symbol"""
```
stlehmann, line 686:
```suggestion
        """Get a single symbol by name."""
```
Please add a period at the end of the docstring.

pyads/ads.py
```diff
@@ -480,6 +483,14 @@ def ams_port(self, port: int) -> None:
             )
         self._adr.port = port
 
+    @property
+    def port(self) -> int:
```
stlehmann, line 487: Please add a docstring.

pyads/ads.py
```diff
@@ -480,6 +483,14 @@ def ams_port(self, port: int) -> None:
             )
         self._adr.port = port
 
+    @property
```
stlehmann, line 486: Why do you want a public port property?

pyads/pyads_ex.py
```diff
@@ -71,7 +71,8 @@
             dll_path += "\\x64"
         dlldir_handle = os.add_dll_directory(dll_path)
     try:
-        _adsDLL = ctypes.WinDLL("TcAdsDll.dll")  # type: ignore
+        # _adsDLL = ctypes.WinDLL("TcAdsDll.dll")  # type: ignore
+        _adsDLL = ctypes.WinDLL("C:/TwinCAT/AdsApi/TcAdsDll/TcAdsDll.dll")
```
stlehmann, line 75: It is never a good idea to hard-code the path to a library. You never know where the library is installed on the system. The purpose of the above line ist that the system finds it itself in the dedicated directories. So please revert this change.

stlehmann (changes requested):

pyads/ads.py
RobertoRoos: This allows `AdsSymbol` (who references to a `Connection` instance) to call the `adsGetSymbolInfo()` function.
Saying this out loud it might be better to create a new method in `Connection` instead and get rid of this `port` property. Changing it now.

pyads/pyads_ex.py
RobertoRoos: Absolutely right, this is a left-over change from my process of the getting the tests to run.

pyads/ads.py
RobertoRoos: Removed those new virtual properties.

pyads/ads.py
```diff
@@ -80,13 +83,12 @@
...
 
 from .structs import (
-    AdsSymbol,
     AmsAddr,
     SAmsNetId,
     AdsVersion,
     NotificationAttrib,
     SAdsNotificationHeader,
-    SAdsSumRequest,
+    SAdsSumRequest
```
stlehmann, line 91: Please don't kill the comma at the end of the line within brackets. It makes version control much easier with them.

pyads/ads.py
```diff
@@ -713,11 +729,16 @@ def get_all_symbols(self) -> List[AdsSymbol]:
                 comment_end_ptr = comment_start_ptr + comment_length
 
                 name = decode_ads(symbol_list_msg[name_start_ptr:name_end_ptr])
-                symtype = decode_ads(symbol_list_msg[type_start_ptr:type_end_ptr])
-                comment = decode_ads(symbol_list_msg[comment_start_ptr:comment_end_ptr])
+                type_name = decode_ads(symbol_list_msg[
```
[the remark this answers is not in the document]
stlehmann: I prefer symbol_type as a name. You can use a variable type-annotation to indicate that it is a string.

pyads/symbol.py
```diff
@@ -0,0 +1,277 @@
...
+    :ivar plc_type: ctypes type of variable (from constants.PLCTYPE_*)
+    :ivar comment: Comment of symbol
+    """
+
+    def __init__(self,
+                 plc: "Connection",
+                 name: Optional[str] = None,
+                 index_group: Optional[int] = None,
+                 index_offset: Optional[int] = None,
+                 type_hint: Optional[Union[Type, str, int]] = None,
```
stlehmann, line 47: Can we not just use type_hint: Optional[str] here? AdsSymbol objects are only created in Connection.get_symbol and Connection.get_all_symbols. The first one doesn't provide type_hint and the second passes a string. So I would rename type_hint to symbol_type and make it Optional[str]. This also makes the following logic easier.

stlehmann (changes requested):

pyads/ads.py
```diff
@@ -526,6 +528,15 @@ def get_local_address(self) -> Optional[AmsAddr]:
...
 
+    def get_symbol_info(self, name: str):
+        """Get information struct of a symbol by its name.
+
+        To create a symbol instance, use Connection.get_symbol() instead.
+        """
+        # Wrap this function inside the class so self._port does not have to
+        # be exposed.
+        return adsGetSymbolInfo(self._port, self._adr, name)
+
```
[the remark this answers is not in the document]
RobertoRoos: Makes sense. Shall I bring back the port and address virtual properties then?

pyads/ads.py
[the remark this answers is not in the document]
RobertoRoos: `symbol_type` as a replacement for `type_hint` or for `type_name`?

The idea was that `type_hint` could be PLCTYPE_*, string or integer (anything that an actual type can be deduced from). I think it's good to emphasize the versatility.

pyads/symbol.py
RobertoRoos: Yeah, let's do that.

At first I was using `info.dataType` (an integer in SymbolInfo) to resolve the type, since resolving an integer would be faster and more robust than matching a string.
Then I noticed this fails for arrays. But let's go all the way then.

pyads/ads.py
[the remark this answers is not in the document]
stlehmann: I think it is sufficient if you call Connection._port and Connection._adr directly.

pyads/symbol.py
```diff
@@ -0,0 +1,267 @@
...
+
+        # Now `self._type_hint` must have a value, find the actual PLCTYPE
+        # from it.
+        # This is relevant for both lookup and full user definition.
+
+        if self.symbol_type is not None:
+            self.plc_type = self.get_type_from_str(self.symbol_type)
+        else:
+            # Set directly from user input
+            self.plc_type = self.symbol_type  # type: Any
```
stlehmann, line 94: So basically you set plc_type to None if no symbol_type was provided, otherwise you try to get the type?

```suggestion
        self.plc_type: Optional[Any] = None
        if self.symbol_type is not None:
            self.plc_type = self.get_type_from_str(self.symbol_type)
```

pyads/symbol.py
```diff
@@ -0,0 +1,267 @@
...
+
+    def make_symbol_from_info(self):
```
stlehmann, line 96: Can we rename this function to create_symbol_from_info to make it consistent with the codebase?

pyads/structs.py
```diff
@@ -327,6 +306,30 @@ class SAdsSymbolEntry(Structure):
...
+                                  representation of the type
+        info.commentLength: 20  # Number of characters in the comment
+        info.stringBuffer: <pyads.structs.c_ubyte_Array_768 object>
+                                # Concatenation of all string info
+        bytes(info.stringBuffer): b'GVL.counter\x00UINT\x00 Counter (in '
+                                  'pulses)\x00\x95\x19\x07\x18\x00\x00\x00\x00'
+        bytes(info.stringBuffer).encode(): "GVL.counter UINT (in pulses)"
+
+        info.name(): "GVL.counter"      # The name section from the buffer
+        info.symbol_type(): "UINT"        # The symbol_type section from the
```
stlehmann, line 330: These are properties, right?

```suggestion
        info.name: "GVL.counter"      # The name section from the buffer
        info.symbol_type: "UINT"        # The symbol_type section from the
```

pyads/structs.py
```diff
@@ -349,17 +352,17 @@ def _get_string(self, offset, length):
 
     @property
     def name(self):
-        "The symbol name"
+        """The symbol name."""
         return self._get_string(0, self.nameLength)
 
     @property
     def type_name(self):
```
stlehmann, line 359:
```suggestion
    def symbol_type(self):
```

pyads/ads.py
```diff
@@ -713,11 +720,16 @@ def get_all_symbols(self) -> List[AdsSymbol]:
                 comment_end_ptr = comment_start_ptr + comment_length
 
                 name = decode_ads(symbol_list_msg[name_start_ptr:name_end_ptr])
-                symtype = decode_ads(symbol_list_msg[type_start_ptr:type_end_ptr])
-                comment = decode_ads(symbol_list_msg[comment_start_ptr:comment_end_ptr])
+                type_hint = decode_ads(symbol_list_msg[
```
stlehmann, line 723:
```suggestion
                symbol_type = decode_ads(symbol_list_msg[
```

pyads/ads.py
```diff
@@ -713,11 +720,16 @@ def get_all_symbols(self) -> List[AdsSymbol]:
...
+                                   type_start_ptr:type_end_ptr])
+                comment = decode_ads(symbol_list_msg[
+                                     comment_start_ptr:comment_end_ptr])
 
                 ptr = ptr + read_length
-                symbol = AdsSymbol(index_group, index_offset, name, symtype, comment)
+                symbol = AdsSymbol(plc=self, name=name,
+                                   index_group=index_group,
+                                   index_offset=index_offset,
+                                   symbol_type=type_hint, comment=comment)
```
stlehmann, line 732:
```suggestion
                                   symbol_type=symbol_type, comment=comment)
```

pyads/symbol.py
```diff
@@ -0,0 +1,267 @@
...
+        self.index_offset = info.iOffs
+        if info.comment:
+            self.comment = info.comment
+
+        # info.dataType is an integer mapping to a type in
+        # constants.ads_type_to_ctype.
+        # However, this type ignores whether the variable is really an array!
+        # So are not going to be using this and instead rely on the textual
+        # type
+        self.symbol_type = info.type_name  # Save the type as string
```
stlehmann, line 114:
```suggestion
        self.symbol_type = info.symbol_type # Save the type as string
```

stlehmann (changes requested):

pyads/structs.py
```diff
@@ -349,17 +353,17 @@ def _get_string(self, offset, length):
 
     @property
     def name(self):
```
stlehmann, line 355: I just see that we don't have any type annotations for the whole module. I need to take care of this. If you like you can already add type annotations to your new properties.

pyads/ads.py
```diff
@@ -713,11 +720,16 @@ def get_all_symbols(self) -> List[AdsSymbol]:
                 comment_end_ptr = comment_start_ptr + comment_length
 
                 name = decode_ads(symbol_list_msg[name_start_ptr:name_end_ptr])
-                symtype = decode_ads(symbol_list_msg[type_start_ptr:type_end_ptr])
-                comment = decode_ads(symbol_list_msg[comment_start_ptr:comment_end_ptr])
+                symbol_type = decode_ads(symbol_list_msg[
+                                   type_start_ptr:type_end_ptr])
+                comment = decode_ads(symbol_list_msg[
+                                     comment_start_ptr:comment_end_ptr])
```
stlehmann, line 726: Multi-line statements should be aligned properly following [PEP8](https://www.python.org/dev/peps/pep-0008/#indentation). In this case putting it on a single line is even more readable.

```suggestion
                symbol_type = decode_ads(symbol_list_msg[type_start_ptr:type_end_ptr])
                comment = decode_ads(symbol_list_msg[comment_start_ptr:comment_end_ptr])
```

pyads/symbol.py
```diff
@@ -0,0 +1,265 @@
...
+    :ivar comment: Comment of symbol
+    """
+
+    def __init__(self,
+                 plc: "Connection",
+                 name: Optional[str] = None,
+                 index_group: Optional[int] = None,
+                 index_offset: Optional[int] = None,
+                 symbol_type: Optional[str] = None,
+                 comment=None):
```
stlehmann, line 45: Return type for __init__ should be None following https://www.python.org/dev/peps/pep-0484/

stlehmann, line 45:
```suggestion
    def __init__(
        self,
        plc: "Connection",
        name: Optional[str] = None,
        index_group: Optional[int] = None,
        index_offset: Optional[int] = None,
        symbol_type: Optional[str] = None,
        comment=None,
    ):
```

pyads/symbol.py
```diff
@@ -0,0 +1,265 @@
...
+    def __del__(self):
+        """Destructor"""
+        self.clear_device_notifications()
+
+    def add_device_notification(
+            self,
+            callback: Callable,
+            attr: Optional[NotificationAttrib] = None,
+            user_handle: Optional[int] = None
+    ) -> Optional[Tuple[int, int]]:
```
stlehmann, line 169:
```suggestion
    def add_device_notification(
        self,
        callback: Callable,
        attr: Optional[NotificationAttrib] = None,
        user_handle: Optional[int] = None,
    ) -> Optional[Tuple[int, int]]:
```

stlehmann: Indentation should always be 4 spaces.

stlehmann: You can just run black against the codebase and it should fix the indentation issues amoung others.

stlehmann (changes requested):
Looks good so far. Mainly small stylish issues remaining.

pyads/structs.py
RobertoRoos: I've added a  few

pyads/symbol.py
RobertoRoos: Yup, I just ran black

pyads/symbol.py
```diff
@@ -0,0 +1,271 @@
...
+            self.comment = info.comment
+
+        # info.dataType is an integer mapping to a type in
+        # constants.ads_type_to_ctype.
+        # However, this type ignores whether the variable is really an array!
+        # So are not going to be using this and instead rely on the textual
+        # type
+        self.symbol_type = info.symbol_type  # Save the type as string
+
+    def read_write_check(self):
```
stlehmann, line 119: We can omit the read_write_check? We should do all checks on initialization. So I see no obvious reason to check again especially as the check is done on every read/write.

pyads/symbol.py
```diff
@@ -0,0 +1,271 @@
...
+            self._handles_list = []  # Clear the list
+
+    def del_device_notification(self, handles: Tuple[int, int]):
+        """Remove a single device notification by handles"""
+        if handles in self._handles_list:
+            self._plc.del_device_notification(*handles)
+            self._handles_list.remove(handles)
+
+    @staticmethod
+    def get_type_from_str(type_str: str) -> Any:
```
stlehmann, line 215: We can make this function private as it is not needed outside this class.

pyads/symbol.py
```diff
@@ -0,0 +1,271 @@
...
+            if scalar_type_str in constants.PLC_ARRAY_MAP:
+                return constants.PLC_ARRAY_MAP[scalar_type_str](size)
+
+        # We allow unmapped types at this point - Instead we will throw  an
+        # error when they are being addressed
+
+        return None
+
+    @staticmethod
+    def get_type_from_int(type_int: int) -> Any:
```
stlehmann, line 263: Is this still needed? As I recall we now use only strings to get the PLCTYPE, right?

stlehmann (changes requested):
Making great progress here. I was just thinking about the handling of reading / writing data and buffering. It might give much speed advantage to buffer the current value (last read) and make it accessible via Symbol.value without any read/write operation. This way one can always access the current value without needing to buffer it in another variable.
So I think we could tweek the API a bit:

Symbol.read() -> Any ... reads value from plc and stores it in Symbol.value. It also returns the value directly for immediate usage.

Symbol.write(new_value: Optional[Any])...Writes new_value to the plc. Also sets Symbol.value to new_value. If new_value is None Symbol.value will be sent to plc.

Symbol.value is a variable which buffers the plc value.

RobertoRoos:
Yeah, the buffering idea sounds good.

One danger would be a user expecting that `symbol.value` is identical to `symbol.read()`, so we have to make very clear the buffering

pyads/ads.py
```diff
@@ -433,7 +434,7 @@ def bytes_from_dict(
     return byte_list
 
 
-class Connection(object):
+class Connection:
```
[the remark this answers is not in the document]
RobertoRoos: It's no longer required past Python 2.7.
Sorry, I didn't think about it much. I realise tackling different problems in one PR is not convenient. Shall I change it back here?

stlehmann:
> so we have to make very clear the first is a buffered value.

If you like you can add a hint to the docstring of value. Also it should be mentioned in the documentation. Anyway I think we should create a separate Quickstart chapter for this new feature.

pyads/symbol.py
```diff
@@ -0,0 +1,271 @@
...
+
+        # Now `self.symbol_type` should have a value, find the actual PLCTYPE
+        # from it.
+        # This is relevant for both lookup and full user definition.
+
+        self.plc_type: Optional[Any] = None
+        if self.symbol_type is not None:
+            self.plc_type = self.get_type_from_str(self.symbol_type)
+
+    def create_symbol_from_info(self) -> None:
```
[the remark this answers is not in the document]
RobertoRoos: I liked putting this in a separate method. I think it keeps the constructor shorter and more clear. Can we keep it as it is?

pyads/symbol.py
RobertoRoos: You have a point. I figured that, because the connection will be owned outside the Symbol object, this could result into confusing problems. For example:

```
plc = ...

with plc:
    # Connection is now opened
    symbol = plc.get_symbol(...)

    # ... lot's of code

# User tries to use the symbol again, but connection has been closed now:
symbol.read()
```

Do we want to catch something like this?

pyads/symbol.py
RobertoRoos: You're right, I'll delete it.

pyads/ads.py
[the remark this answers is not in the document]
stlehmann: Yes, please 🙇  If you like you can put it in a separate PR.

pyads/symbol.py
RobertoRoos: Just checked this by the way. WIthout the `read_write_check`, the Connection quietly returns `None` when it has been closed. This might lead to sneaky bugs in users code.

Thoughts?

stlehmann (commented):
This looks very nice now. Only some minor issues and we are ready to merge. If you like you might want to add a chapter to the quickstart guide concering symbols. And the changelog of course :)

pyads/symbol.py
stlehmann: Hm, I actually prefer if Connection throws the exception but this breaks compatibility and might cause issues in some use-cases. So maybe we just check for plc.open on read/write and throw an exception if it is not open. What do you think)

RobertoRoos: I agree. So we check `plc.is_open` on every read and write then?

Maybe we can add an `__init__` keyword argument to `Connection` for compatibility. Something like `read_write_error=False` that determines whether errors are thrown.

pyads/constants.py
```diff
@@ -110,6 +110,11 @@
 
 def PLCTYPE_ARR_REAL(n: int) -> Type[Array]:
     """Return an array with n float values."""
+
+    print()
+    print('I got called!')
+    print()
+
```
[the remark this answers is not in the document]
RobertoRoos: Yikes, why does this always happen 😅

pyads/symbol.py
stlehmann: > Maybe we can add an __init__ keyword argument to Connection for compatibility. Something like read_write_error=False that determines whether errors are thrown.

That would be an option but as Connection is the parent of Symbol there is no way to force that a Connection object is created with this flag set. I think we get the least trade-off if we just check for plc.is_open on each read/write. This is not much computation anyway.

doc/quickstart.md
```diff
@@ -101,6 +101,93 @@ the route to the testserver needs to be added from another python console.
...
+```
+
+The address and type of the symbol will be automatically determined using a READ_WRITE request to the ADS server, based on the variable name. This lookup is skipped when all the information has already been provided:
+
+```python
+>>> import pyads
+>>> plc = pyads.Connection('127.0.0.1.1.1', pyads.PORT_SPS1)
+>>> plc.open()
+# Identical to `plc.get_symbol('global.bool_value')`, remaining info will be looked up:
+>>> symbol = pyads.AdsSymbol(plc, 'global.bool_value')
```
stlehmann, line 124: I wonder what the exact difference is between plc.get_symbol and pyads.AdsSymbol(plc, ...). They seem to do two methods for the same task at least from a user-point-of-view I think. Could we just expand plc.get_symbol to also support passing  all informations and encourage plc.get_symbol?

stlehmann (commented):
Just wondering if we could introduce a device_notification parameter on symbol creation that (if set to true) automatically creates a device notification which changes the buffered value as soon as the plc variable is changed. This way value is always the actual value from the plc.

doc/quickstart.md
RobertoRoos: > Just wondering if we could introduce a device_notification parameter on symbol creation that (if set to true) automatically creates a device notification which changes the buffered value as soon as the plc variable is changed. This way value is always the actual value from the plc.

If only the buffer changes without triggering anything else, what is the point? With such a default callback, `symbol.value` will again be almost identical to `symbol.read()`. If a user wants the most recent value, they can call `.read()` directly.

RobertoRoos: > I wonder what the exact difference is between plc.get_symbol and pyads.AdsSymbol(plc, ...). They seem to do two methods for the same task at least from a user-point-of-view I think. Could we just expand plc.get_symbol to also support passing all informations and encourage plc.get_symbol?

There is not real difference now. We could advocate the use of either the factory method or direct object initialization, and not the other. Does either have your preference?

stlehmann: > If only the buffer changes without triggering anything else, what is the point? With such a default callback, `symbol.value` will again be almost identical to `symbol.read()`. If a user wants the most recent value, they can call `.read()` directly.

I think of a use-case like a plot where the client plots plc values periodically like every 100ms or so. You basically read the current value from value and add it to the plot. With the internal device notification you don't need to poll from the plc every cycle. Instead only changed values will be pushed to the buffer and can be read much faster into the graph.

stlehmann: > There is not real difference now. We could advocate the use of either the factory method or direct object initialization, and not the other. Does either have your preference?

I prefer the factory function as this is more intuitive for me because we use "plc." for everything anyway. But I leave this decisision up to you. If you like to use object initializiation that's fine. But as you say we should just advocate one way of doing it in the Quickstart.

RobertoRoos: I updated the `plc.get_symbol()` method to match with the `AdsSymbol` constructor. I also removed the 'manual' object creation, so only the factory method is used.

I see your point about auto-updating the buffer, so I added such a feature.
I felt like the constructor already had enough arguments, so I've added a method to enable auto-update. Thoughts?

klauer:
Nice feature - this symbol class will definitely be my preferred method of using pyads 👍 

A quick summary of some thoughts from up here in the peanut gallery:
* Consider adding context manager support to `AdsSymbol` (mirroring that of `Connection`), such that a manual `.close()` is not necessary
* `AdsSymbol.__del__` may not be throwing errors for you now, but from experience with other projects, I can say it's a common source of errors when the interpreter is exiting and has already cleaned up objects your `__del__` method relies on. You might consider using [`weakref.finalize`](https://docs.python.org/3/library/weakref.html#weakref.finalize) instead.
* Moving the regular expression-based "type determination from string" utilities out of the `AdsSymbol` class would make testing easier and the functions themselves reusable. Also, `re.compile` is advisable when possible.
* `enable_auto_update(True/False)` is a bit awkward in name - might `set_auto_update` be better?

Apologies if any of this has already been discussed above - I haven't followed this as closely as I'd like to have.

RobertoRoos:
> * Consider adding context manager support to `AdsSymbol` (mirroring that of `Connection`), such that a manual `.close()` is not necessary

How would the context usage look like? The symbol don't really open or close themselves.

I was about to say 'Connection doesn't required .close()', but I know notice it doesn't have its own destructor. Should it?

> * `AdsSymbol.__del__` may not be throwing errors for you now, but from experience with other projects, I can say it's a common source of errors when the interpreter is exiting and has already cleaned up objects your `__del__` method relies on. You might consider using [`weakref.finalize`](https://docs.python.org/3/library/weakref.html#weakref.finalize) instead.

Yeah, I see your point. Is there a downside to using a weakref?  
And would a Symbol keep a weakref to the Connection, or the other way around?

> * Moving the regular expression-based "type determination from string" utilities out of the `AdsSymbol` class would make testing easier and the functions themselves reusable. Also, `re.compile` is advisable when possible.

I think I made the method static before. Would that work?

> * `enable_auto_update(True/False)` is a bit awkward in name - might `set_auto_update` be better?

Agreed, changed it.

doc/quickstart.md
stlehmann: This looks very nice. The naming `set_auto_update` is fitting I think. Anyway I think we could add one more parameter `auto_update=False` to `AdsSymbol.__init__`. By default it will be false but if someone likes to enable it it would be just a breathe to add `auto_update=True`. Especially if one creates multiple symbols I think it would be very convenient to have this handy parameter on initialization instead of writing an extra line for each symbol.

klauer:
> How would the context usage look like? The symbol don't really open or close themselves.

True - I mistakenly was thinking of the `__del__` as an equivalent to `close()` and didn't express that very well. That said, do you think a general `cleanup()` method would be appropriate? In that case, the context manager would from a very high level be "create symbol; clean up notifications (and anything else) when the context block is done".

> I think I made the method [for "type determination from string"] static before. Would that work?
Making it static would be an improvement to not require an instantiated symbol, for sure. I might argue that its functionality is independent of the `AdsSymbol` and should remain a separate utility function, but it could just be a matter of preference at that point. 😄 

>  Is there a downside to using a weakref? And would a Symbol keep a weakref to the Connection, or the other way around?

Since pyads is Python 3.6+, I don't think there would be a downside. Removing `__del__` may even help with eventual PyPy support. For what it's worth, I haven't found a better solution at least.

Keep in mind the important note from the `weakref.finalize` docs (below [atexit](https://docs.python.org/3/library/weakref.html#weakref.finalize.atexit)). It's a bit confusing at first, but to get around that, you'd rather be looking to make a weakref finalizer of `AdsSymbol` that uses the pertinent cleanup information. As the docs note, making it a non-bound method of the form `def _symbol_cleanup(plc, handles_list):` with a stashed finalizer instance like `self._finalizer = weakref.finalize(self, _symbol_cleanup, self.plc, self._handles_list)` might work, I think.

doc/quickstart.md
RobertoRoos: Alright, that can't hurt. I've added it.

RobertoRoos:
Forgive me for being a bit slow in all this, but I don't see how `weakref.finalize` would solve our dependencies problem.

My interpretation: `weakref.finalize` is an alternative to `__del__` methods. So if we add a cleanup method to the symbols, the user can still close or delete the `Connection` object before the symbols are cleaned up.

stlehmann:
Thanks for adding auto_update to AdsSymbol. Subsequently it will also be needed in the get_symbol function :scream:

RobertoRoos:
> Thanks for adding auto_update to AdsSymbol. Subsequently it will also be needed in the get_symbol function 😱

There will come a day that remember to do everything. But it is not this day.

Updated it

pyads/symbol.py
```diff
@@ -0,0 +1,290 @@
...
+        """Remove all registered notifications"""
+        if self._handles_list:
+            for handles in self._handles_list:
+                self._plc.del_device_notification(*handles)
+            self._handles_list = []  # Clear the list
+
+        self._auto_update_handle = None  # If auto-update was enabled,
+        # it won't work anymore
+
+    def del_device_notification(self, handles: Tuple[int, int]):
```
stlehmann, line 213:
```suggestion
    def del_device_notification(self, handles: Tuple[int, int]) -> None:
```

pyads/symbol.py
```diff
@@ -0,0 +1,290 @@
...
+        """
+        if auto_update and self._auto_update_handle is None:
+            self._auto_update_handle = self.add_device_notification(
+                self._value_callback
+            )
+        elif not auto_update and self._auto_update_handle is not None:
+            self.del_device_notification(self._auto_update_handle)
+            self._auto_update_handle = None
+
+    def _value_callback(self, notification: Any, data_name: Any):
```
stlehmann, line 236:
```suggestion
    def _value_callback(self, notification: Any, data_name: Any) -> None:
```

stlehmann (commented):
Nice. Just some minor style and type-annotation issues

stlehmann (changes requested):
Great work. I think this will be a fantastic new feature. Just one small issue concerning the debug-level and we are done.

pyads/testserver.py
```diff
@@ -29,23 +29,28 @@
...
-from collections import namedtuple, defaultdict
+from collections import namedtuple
 
 from pyads import constants
 
 # Log to stdout by default
 logger = logging.getLogger(__name__)
 formatter = logging.Formatter("%(levelname)s:%(message)s")
 stdout_handler = logging.StreamHandler()
-stdout_handler.setLevel(logging.DEBUG)
```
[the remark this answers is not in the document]
RobertoRoos: Yeah, I tried to disable all the connection messages to help find the test messages. Of course I forgot to remove it again...

stlehmann (approved):
@RobertoRoos great work and again thanks for your massive effort towards adding this feature. I already look forward on using it for my next PLC project.

stlehmann:
I merged this PR and released a new version 3.3.1 of pyads.

Issue #183, opened 2021-01-12
Support structured ADS types with read_list_by_name


pylipp:
In a [different PR](https://github.com/stlehmann/pyads/pull/161#issuecomment-720324694) I wondered whether `Connection.read_list_by_name` is compatible with structured ADS variables? E.g.

```
>>> structure_def = (
...    ('rVar', pyads.PLCTYPE_LREAL, 1),
...    ('rVar2', pyads.PLCTYPE_REAL, 1),
...    ('iVar', pyads.PLCTYPE_INT, 1),
...    ('iVar2', pyads.PLCTYPE_DINT, 3),
...    ('sVar', pyads.PLCTYPE_STRING, 1)
... )
>>> plc.read_list_by_name(['global.sample_structure'])
```

@bdcoyle then said it's not, so I'd like to know whether I can assist here.

pylipp:
I think [this](https://infosys.beckhoff.de/english.php?content=../content/1033/tcsample_vc/html/TcAdsDll_API_CPP_Sample17.htm&id=) is a relevant resource but I cannot make out whether sum-read is compatible with structured datatypes.

stlehmann:
I don't think sum read/write is supported for structurer datatypes at the moment. Am I right @bdcoyle?

stlehmann:
I don't know if it is even supported by ADS itself.

bdcoyle:
That's correct - at the moment it's not supported.

I _think_ it should be possible - I assume a structure has an index group and index offset like any other variable in the PLC, these plus the structure size is what is needed to make the sum read/write requests.

A good point to start would be to call adsGetSymbolInfo on a structure in the PLC, and make sure the idx grp and idx offset are valid.

Then within sumRead and sumWrite, an extra case would need to be added to handle structures - it would need to unpack from the binary data returned by the PLC into the structure_def.
Maybe some of the existing functions for handling structures could be used here.

I don't have time at the moment to contribute on this, so feel free to give it a shot!

stlehmann:
Thanks to @pylipp we now also have this feature. I think I can close this issue as it is solved by 
#187

PR #187, opened 2021-01-20
Support structured variables in read_list_by_name

pylipp:
Acc. to discussion in #183.

For support in `write_list_by_name` I'd open a separate PR.

I could not test it with a PLC yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## 3.3.4 [unreleased]
 
 ### Added
+* [#187](https://github.com/stlehmann/pyads/pull/187) Support structured data types in `read_list_by_name`
 * [#195](https://github.com/stlehmann/pyads/pull/195) Read/write by name without passing the datatype
 * [#200](https://github.com/stlehmann/pyads/pull/200) Split read write by list into max-ads-sub-comands chunks
 * [#206](https://github.com/stlehmann/pyads/pull/206) AdsSymbol now supports DT, DATE_TIME and TIME datatypes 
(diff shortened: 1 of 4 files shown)


pylipp:
The unittest failing on Travis locally passes; not sure how to do more inspection here.

chrisbeardy:
The approach of having to pass  the structure definitions in and use it to unpack the bytes is the first approach I pondered about when the question first came up in #183. As previosuly mentioned I don't think ADS has native support for structures as the symbol info just tells you it's a structure, (which is how the read_struct_by_name code came about and annoyingly having to know the structure definition up front), so to me this looks like a good "workaround" to allow you to mix the read_list and read_struct fucntionality. I'm quite busy on projects at the minute but I may be able to test at some point.

However if you do have the time, you do not neccesarily need a PLC to test it. If you have a windows laptop or can spin up a Windows virtual machine, you can install TwinCAT 3 on that and test it by pointing the target system to local / or the VM.

stlehmann (commented):
@pylipp This PR looks good to me. I also tried a sample on my PC which worked well.

stlehmann:
I don't know why the CR fails, though. It might be some issue related to adslib which is used on Linux only.

pylipp:
Will update this after the changes of #200 and #202 !

chrisbeardy:
> 
> 
> Will update this after the changes of #200 and #202 !

sorry, I knew that #200 would mean that this would need rewoking. But I thought it to be a sensible addition to add in.

pylipp:
> > Will update this after the changes of #200 and #202 !
> 
> sorry, I knew that #200 would mean that this would need rewoking. But I thought it to be a sensible addition to add in.

Totally! Was straightforward to update the present branch :)

stlehmann (approved):

stlehmann:
Great work @pylipp 👍. I'll merge it right away.

PR #192, opened 2021-02-10
Make AdsSymbol even more pythonic

stlehmann:
The purpose of this PR is to make the great AdsSymbol class even more Pythonic by:

* replacing setter/getter functions by properties
* Allowing immediate writing of the value to the plc if auto_update is True

### DIFF
--- a/pyads/symbol.py
+++ b/pyads/symbol.py
@@ -308,6 +308,7 @@ class AdsSymbol:
 
         """
         self._value = val
+
         # write value to plc if auto_update is enabled
-        if self._auto_update_handle is not None:
+        if self.auto_update:
             self.write(val)
--- a/pyads/symbol.py
+++ b/pyads/symbol.py
@@ -8,9 +8,10 @@ the circular dependencies.
 import re
 from ctypes import sizeof
 from typing import TYPE_CHECKING, Any, Optional, List, Tuple, Callable
+
+from . import constants  # To access all constants, use package notation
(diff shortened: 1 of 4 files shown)


stlehmann:
@RobertoRoos I made some minor API-changes to make AdsSymbol class more pythonic. You might want to have a look at it and tell me if you approve it.

pyads/symbol.py
```diff
@@ -288,3 +270,44 @@ def get_type_from_str(type_str: str) -> Any:
...
+    @value.setter
+    def value(self, val: Any) -> None:
+        """Set the current value of the symbol.
+
+        If auto_update is True then the the write command will be called automatically.
+
+        """
+        self._value = val
+        # write value to plc if auto_update is enabled
+        if self._auto_update_handle is not None:
```
RobertoRoos, line 312: I think `if self.auto_update` it's a little easier to read.

RobertoRoos:
I am personally not the biggest fan of magic properties, but I understand this is considered good practice.

Placed a teeny-tiny comment, I like your changes. Thank you for asking me.

pyads/symbol.py
stlehmann: That's a good suggestion. I implemented the auto_update property after the value property so I didn' t see this. I will change it.

stlehmann:
@RobertoRoos thanks for having a quick look at it :)

PR #200, opened 2021-02-15
Split read write by list into max-ads-sub-comands chunks

chrisbeardy:


chrisbeardy:
The tests are failing for the new sub-commands tests, they are failing with a:

```
pyads.pyads_ex.ADSError: ADSError: parameter size not correct (1797).
```

I submitted the PR still as I can't work out why the testserver is returning this, so would appreciate some help? I have tested with an actual PLC, and the new code works as expected but just not in the test with the testserver. The spliting of lists function works so what is actually being passed into adsSumRead is what is expected so I can only see that there is an issue with the testserver not responding as expected?

The code I used to test with the PLC is here:

I just set up an array in MAIN:

```
iTest: ARRAY [0..1400] OF INT;
```

```python
import pyads
import time

plc = pyads.Connection("169.254.198.195.1.1", 851)

plc.open()


now = time.time()
test_array = {f"MAIN.iTest[{i}]": i for i in range(0, 1401)}
plc.write_list_by_name(test_array)
print(plc.read_list_by_name(test_array))
test_array = {f"MAIN.iTest[{i}]": 0 for i in range(0, 1401)}
plc.write_list_by_name(test_array)
print("First Written: " + str(time.time() - now))
test_array = {f"MAIN.iTest[{i}]": i for i in range(0, 1401)}
plc.write_list_by_name(test_array)
test_array = {f"MAIN.iTest[{i}]": 0 for i in range(0, 1401)}
plc.write_list_by_name(test_array)
print("Second Written: " + str(time.time() - now))
test_array = {f"MAIN.iTest[{i}]": i for i in range(0, 1401)}
plc.write_list_by_name(test_array, ads_sub_commands=1401)
test_array = {f"MAIN.iTest[{i}]": 0 for i in range(0, 1401)}
plc.write_list_by_name(test_array, ads_sub_commands=1401)
print("Third Written: " + str(time.time() - now))
test_array = {f"MAIN.iTest[{i}]": i for i in range(0, 1401)}
plc.write_list_by_name(test_array, ads_sub_commands=2)
test_array = {f"MAIN.iTest[{i}]": 0 for i in range(0, 1401)}
plc.write_list_by_name(test_array, ads_sub_commands=2)
print("Fourth Written: " + str(time.time() - now))


test_array = [f"MAIN.iTest[{i}]" for i in range(0, 1400)]

print(plc.read_list_by_name(test_array, ads_sub_commands=1))
print("Second Read: " + str(time.time() - now))
plc.close()
```

chrisbeardy:
parameter size not correct suggests that the test server is not correctly storing the size between reads/writes of the test variables?

I don't know if @pylipp or @bdcoyle have any thoughts? I can't remember who added the intial read/write list functionailty?

stlehmann:
That really seems to be a bug or limitation of the testserver. It can only handle 3 or 4 variables on AdsSumRead. 

1 variable: Warning: Frame too long: 31>17
2 variables: Frame too long: 31>26
3 variables: OK
4 variables: OK
5 variables: KeyError: 1702101505
6 variables: KeyError: 1702101505

stlehmann:
I actually don't know what's happening here exactly but it seems the SUMUP_READ is limited to return 3 values: 1, 2 and "test". So you might want to adjust your test to read 6 variables and split by 3.

https://github.com/stlehmann/pyads/blob/0c93dfe09e0ab295d144745427813b367e75598b/pyads/testserver.py#L494-L499

chrisbeardy:
Ok, I'll give that a go

bdcoyle:
Sorry for the slow response, I remember when writing the tests for the sum read/write I was also getting the 'Warning: Frame Too Long'. It seems to come from AdsLib (AmsConnection.cpp) but I couldn't work out why.
Since it was working when I tested on an actual PLC, I made the hardcoded response within the limit as mentioned above for the purpose of the testserver

https://github.com/stlehmann/ADS/blob/d0804bc4edc68851c75f982e0aebc7c2386b0a29/AdsLib/AmsConnection.cpp#L244-L248

stlehmann:
@chrisbeardy I'll take a deeper dive in the sum read handling on the testserver and will try to make it work with variable length and datatypes. So you might not need to change your tests at all.

chrisbeardy:
@stlehmann that would be great if possible

stlehmann:
@chrisbeardy thanks for this PR and sorry for the mess with the commits. Actually I wanted to rebase on master but this didn't work out so well.

tests/test_ads.py
```diff
@@ -561,6 +561,61 @@ def test_bytes_from_dict(self):
         with self.assertRaises(KeyError):
             pyads.bytes_from_dict(OrderedDict(), structure_def)
 
+    def test_dict_slice_generator(self):
```
stlehmann, line 564: Can we remove the test slice generators as they are not used?

stlehmann (changes requested):

chrisbeardy:
I always struggle with rebasing a PR in GitHub! Half the time I end up creating a new PR. Thanks for the great work on pyads and others on some of the latest features. It's allowing me to build a really powerful front end for a machine I'm working on now which would have been not as easy otherwise.

stlehmann:
The rebase-button of PyCharm was so tempting 😞

tests/test_ads.py
chrisbeardy: do you mean just remove the tests? as the functons are still used in the read/write_list_by name functions in ads.py.

chrisbeardy: even though the tests for the read/write_list `test_read_list_ads_sub_commands` technically cover the usage of the _slice_generators, I would argue as they are standalone functions, it is still important to test them with tests spefic to test those standalone functions?

stlehmann: Ah sorry I didn't see that these are tests. I thought it to be a generator for some test data so I looked if they are called anywhere in the code which was not the case. Alright then all is set and I shall merge this PR :)

chrisbeardy:
> 
> 
> @chrisbeardy thanks for this PR and sorry for the mess with the commits. Actually I wanted to rebase on master but this didn't work out so well.

Actually when you merge a PR in GitHub, do you get the option to rebase it then?

stlehmann:
Might be. I think the problem was that I did a rebase in between.

PR #209, opened 2021-02-17
removed duplicate tests and added addtional asserts to existing tests

chrisbeardy:
removed some duplicate tests. read/write_list_sub_commands had their tests duplicaed, possibly casused by merge of last PR. Also added more test crtieria for the no cache tests  on rad and write_list_by_name to make sure that the symbol is not cached.


stlehmann (approved):
Thanks for fixing this. Seems the rebase was not a good idea alltogether. I'll be more careful with that in the future.

PR #213, opened 2021-02-18
Support for structure types in write_list_by_name

pylipp:
As follow-up to #187, I'd like to add support for structured types in `write_list_by_name`.

The implementation should be quite straightforward, yet the unit test fails when I try to verify the written data by reading it again. Or could it be a limitation of the testserver?

### DIFF
--- a/pyads/ads.py
+++ b/pyads/ads.py
@@ -915,6 +915,7 @@ class Connection(object):
         data_names_and_values: Dict[str, Any],
         cache_symbol_info: bool = True,
         ads_sub_commands: int = MAX_ADS_SUB_COMMANDS,
+        structure_defs: Optional[Dict[str, StructureDef]] = None,
     ) -> Dict[str, ADSError]:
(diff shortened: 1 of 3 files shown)


chrisbeardy:
I believe this is a limititation of the test server, I learnt recently that the test server on read is set ony to return given values, so is more used to validate the read happened. On write, if the write happends you get back the reposnse you are testing for, i.e. "no error", I dont believe it keeps in memory any values written to it. The tests therefore make the assumption that if the test server says the write happened, it happened. Or at least this is the case for other write tests.

stlehmann:
As @chrisbeardy already wrote the TestServer is very basic. There is an AdvancedHandler that allows to store values and return them on read operations. But this only supports basic datatypes.

stlehmann:
Sorry it seems I just merged that PR before proper review. To prevent further confusion I'll keep it in master and fix the tests in a separate PR. Thanks @pylipp for the nice work.

Issue #201, opened 2021-02-17
Simplify Type Handling


koppa:
Currently are there are multiple different ways how to pass a type to the pyads module:

* As Strings:
   * "LREAL"
   * "ARRAY[0..10] of LREAL"
* pyads.PLCTYPE_X
* ctypes type (Structures, etc...)
* "string" for the struct module

I would propose to unify this handling, that every method accepts the same typedefintions.
For example the AdsSymbol has implemented its own parsing code.

Also related to: #189

RobertoRoos:
I agree. We could pool all type interpretations into a static function that will either produce a ctype-like type, read to be used by Ads or raises an exception.
This function can then be used by each read and write function, and by AdsSymbol.

chrisbeardy:
As far as the public API, I see 3 types ways of defining types available?:

- For read/write_by_name,, read/write_structure_by_name, using pyads.PLCTYPE_*
- For AdsSymbol, using  a string, "BOOL"
- Notification callbacks using ctypes.Structures
 
@koppa  What is the "string" for the struct module?

I can see why each method was used to acheive its desired functionality. But I can also need a need to unify this.

## Unifying the type info for the user

I would argue that the ctypes.Structure could be left as is, it seems quite a standalone use case? and possibly superseeded by the structure def approach which we could point new users too?:

![image](https://user-images.githubusercontent.com/20585410/108178330-3c429c80-70fc-11eb-8059-964bb1bf658d.png)

The original way from read/wirte_by_name was using pyads.PLCTYPE* and then if you needed an array, multiplying it by array size:

![image](https://user-images.githubusercontent.com/20585410/108179162-326d6900-70fd-11eb-8158-8d9ee65d2a25.png)

however since the introductions of read/write_list_by_name and the ADSSymbol pyads seems to be moving toward a more finding the type and caching the symbol address etc, which makes a lot of sense and abstracts a lot of the detail from the user.  For unifying the ADS_SYMBOL and others, I see currently not too great a harm leaving it as it right now for two reasons.

1) It will bring braking changes, not that that is a big issue, but I think right *now* its not a big enough problem to refactor it.
2) I *believe* for a majority of use cases going forward (especially now #195 has been merged too) one will not provide plc type info (except for write/read_structure_by_name), as pyads will get the info itself , so I see it more of an advanced feature for speeding up your code or for edge cases, perhaps such as #189.

[...]

PR #214, opened 2021-02-19
Added PLCTYPE_* as the prefered type of symbol initialization

RobertoRoos:
Title says it all. Should solve #201 .

Two questions:

- What is the correct type hint for a ctypes class, like the `pyads.PLCTYPE_*` constants?
- Should be also modify `Connection.read` and `.write` to accept the same arguments as `get_symbol`, i.e. both a ctypes class and string?

### DIFF
--- a/doc/documentation/symbols.rst
+++ b/doc/documentation/symbols.rst
@@ -28,10 +28,11 @@ provided:
    >>> symbol = plc.get_symbol('global.bool_value')
    # Alternatively, specify all information and no lookup will be done:
    >>> symbol = plc.get_symbol('global.bool_value', index_group=123,
-                               index_offset=12345, symbol_type='BOOL')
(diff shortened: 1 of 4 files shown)


stlehmann (approved):

stlehmann:
> What is the correct type hint for a ctypes class, like the pyads.PLCTYPE_* constants?

As all PLCTYPES are ctypes the correct type would be PyCSimpleType. I suggest we create a type alias PLCSimpleType for that.

    PLCSimpleType = PyCSimpleType

> Should be also modify Connection.read and .write to accept the same arguments as get_symbol, i.e. both a ctypes class and string?

I would not do that. We want to encourage usage of PLCTYPE_ constants for the reasons named in #201 so adding string support for these functions would lead us in the other direction and add overhead to the codebase.

stlehmann:
Thanks @RobertoRoos for re...re...implementing this feature and sorry again for having you remove it in the first place 🙏

PR #215, opened 2021-02-19
Add datatype annotations

stlehmann:
This PR introduces PLCSimpleDataType and PLCDataType to make type-annotations more specific. 

Also it renames the symbol_type parameter of Connection.get_symbol to plc_datatype to fit in the pattern of functions like read or read_by_name. I hope @RobertoRoos doesn't mind this small change.

### DIFF
--- a/pyads/ads.py
+++ b/pyads/ads.py
@@ -613,7 +613,7 @@ class Connection(object):
         return None
 
     def write(
-        self, index_group: int, index_offset: int, value: Any, plc_datatype: Type[PLCDataType]
+        self, index_group: int, index_offset: int, value: Any, plc_datatype: Type["PLCDataType"]
     ) -> None:
         """Send data synchronous to an ADS-device.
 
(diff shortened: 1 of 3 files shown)


RobertoRoos:
Looks good. Do the new renders properly in Sphinx documentation?

stlehmann:
> Do the new renders properly in Sphinx documentation?

grafik

stlehmann:
Does anyone know a way how to tell Sphinx to not solve the Type aliases? I think that is a lot of verbose information here.

RobertoRoos:
I'll say that's not terrible clear. Can we let Sphinx not resolve the type?

stlehmann:
> Couldn't we use something like Type[_CData] instead?

Actually I was thinking about this in the first place. All basic ctypes types are _ctypes.PyCSimpleType. But I couldn't find
the module/package from which to import it.

stlehmann:
Putting the Type in quotation marks keeps Sphinx from expanding the type information. Looks good to me. What do you think?

RobertoRoos:
Looks fine, good work!

stlehmann:
Maybe in time someone comes up with a better idea then just stacking all basic ctypes together in PLCSimpleDataType I'm not too happy with the current solution but don't see a better way at the moment.

PR #218, opened 2021-02-21
Support structured data types in write_list_by_name

stlehmann:
> As follow-up to #187, I'd like to add support for structured types in write_list_by_name.
The implementation should be quite straightforward, yet the unit test fails when I try to verify the written data by reading it again. Or could it be a limitation of the testserver?


stlehmann:
Sorry, I had to recreate the PR because I messed up the original. Please see #213 for original conversation

PR #220, opened 2021-02-21
Structure support for write list by name

stlehmann:
This is rebase. The original one can be found here: #213


stlehmann:
The test still fails apparently because AdvancedHandler does not support structures at the moment. I'll see to that.

stlehmann:
This PR also introduces sumup_read and sumup_write support on the AdvancedHandler of the testserver.

Issue #207, opened 2021-02-17
AdsSymbol: add full support for structures


stlehmann:
At the moment AdsSymbols don't support structures. 

The nicest way I can think of would be to read a structure definition via some ADS command? But I doubt that there exists such a command. 

So maybe we can use the approach of providing a structure definition as in read_structure_by_name? @RobertoRoos what are your thoughts on this?

RobertoRoos:
Yeah from what I've seen in my tests this afternoon, you cannot get the complete struct definition over ADS. So the user will have to provide it.

Users could simply provide it with the `symbol_type` argument. We can capture a Tuple type (which should never be passed besides structs) and set an internal flag `_is_struct = True`.

In read()/write() we can add an if-else to use the `read_structure_by_name()` in case `_is_struct == True`.

We can keep the returned value identical to what's returned by `read_structure_by_name()`, which is an `OrderedDict` I think.

stlehmann:
I like the idea to include read_structure_by_name. That keeps the codebase small.

> symbol.value[field] = 3.14

Oh that's tricky. I think in this case the wrapper would return an object and __setitem__ would be called. This mean the getter needs to return a wrapper object that uses __setitem__ to update the target value. I don't think we should support that. Maybe we can handle structs more the "tuple-way":

```python
   struct = symbol.value
   struct["a"] = 123
   symbol.value = struct
```

PR #223, opened 2021-03-01
Add structure support to AdsSymbol class

stlehmann:
This addresses #207 .

It adds structure_def and array_size parameters to Connection.get_symbol. This way symbols can be created for structures and arrays of structures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## 3.3.5 [unreleased]
 
 ### Added
+* [#223](https://github.com/stlehmann/pyads/pull/223) Add structure support for symbols
 
 ### Changed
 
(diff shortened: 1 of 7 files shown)


stlehmann:
@RobertoRoos would you mind to give this PR a quick review?

RobertoRoos (approved):
Looks good @stlehmann , no real comments.

I have to say a part of me dislikes these features-on-top-features types of code, with if-else to steer it's behavior.
This is not entirely rational, and for now it should definitely be fine. If in the future more big features are added, we should consider creating symbol sub-classes. A struct-symbol for example could be its own class.
But again, good for now.
Good work!

stlehmann:
> I have to say a part of me dislikes these features-on-top-features types of code, with if-else to steer it's behavior.
This is not entirely rational, and for now it should definitely be fine. If in the future more big features are added, we should consider creating symbol sub-classes. A struct-symbol for example could be its own class.

I am definetly with you on that. Maybe we can tackle this  in some future PRs.

Issue #224, opened 2021-03-02
AdsSymbol returns Bytes not STRING


chrisbeardy:
I'm trying to use ADSSymbol to read/write a string but I'm getting TypeError if I pass a string to write and when i read I get a list of bytes and not the string representation. Is this intended behaviour?

PLC declaration:
![image](https://user-images.githubusercontent.com/20585410/109670015-9dda2080-7b6a-11eb-8279-121018435809.png)

![image](https://user-images.githubusercontent.com/20585410/109670088-afbbc380-7b6a-11eb-8a89-b28dc0fe9b64.png)

![image](https://user-images.githubusercontent.com/20585410/109670172-c19d6680-7b6a-11eb-9e75-81c7959126a4.png)

chrisbeardy:
doing this in my end code sorts it our for me but just wondering if this is expected?

```python
self._display_plc_symbol_type = self._display_plc_symbol.symbol_type
if "STRING" in self._set_plc_symbol_type:
    value = value.encode()
self._set_plc_symbol.write(value)

value = self._display_plc_symbol.read()
if "STRING" in self._display_plc_symbol_type:
    value = value.decode()
```

chrisbeardy:
UPDATE: this also seems to affect PR #195 trying to read a string using `read_by_name` and not pasisng in datatype now also reads a bytes list and not a string.

![image](https://user-images.githubusercontent.com/20585410/109805988-07663780-7c1c-11eb-94fe-292e78243c40.png)

chrisbeardy:
it appears to recognise string stypes as PyCArrayTypes when using adsGetSymbolInfo thereforedoes not do the encoding.

![image](https://user-images.githubusercontent.com/20585410/109807600-06360a00-7c1e-11eb-9bda-0c17fef3ab2c.png)

Should  pyads.ex be modified, There are a few places that **may** need modification. `adsSyncReadReqEx2`, `adsSyncWriteReqEx` and `adsSyncReadWriteReq2" seem to be the main ones but there are loads of places where a check is made against PLCTYPE_STRING to determine what to do with the response and encode/decode the bytes. 

e.g.
![image](https://user-images.githubusercontent.com/20585410/109808361-e8b57000-7c1e-11eb-8e02-fff54c64cfe4.png)

Not sure wht the knock on effects to AdsSymbol would be (or anything else) here if we modified that a PyCArrayType is also encoded/decoded, or should this be done further up the tree in the callin functions, e.g. read_by_name

[...]

PR #225, opened 2021-03-04
Added proper capture of char arrays

RobertoRoos:
If a data_type was passed as a char array, it wouldn't get recognized as a string.

Tackles #224

### DIFF
--- a/pyads/pyads_ex.py
+++ b/pyads/pyads_ex.py
@@ -197,6 +197,21 @@ def send_raw_udp_message(
         return sock.recvfrom(expected_return_length)
 
 
+def type_is_string(plc_type: Type) -> bool:
+    """Return true if the given class is a string type."""
+
+    # If single char
+    if plc_type == PLCTYPE_STRING:
+        return True
+
+    # If char array
+    if type(plc_type).__name__ == "PyCArrayType":
+        if plc_type._type_ == PLCTYPE_STRING:
+            return True
+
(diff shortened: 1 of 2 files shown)


chrisbeardy:
thanks for looking into this so quickly, I did not reserve the time to look deeper the other day. I've just had a qucik look through and I think this seems sensible, as you say in the issue ticket get symbol correctly deduces thats its a string but it's the pyads_ex functions that are not recognising it as they are looking for you to pass the PLCTYPE_STRING in as a datatype. If you dont pass in the STRING type they can't figure out the char array is a string and dont do the encoding/decoding.

 Out of interest do you know what the PyCArrayType is actually used for? When would we actually hit this line in say `adsSyncReadWriteReqEx2`?
image

RobertoRoos:
`PyCArrayType` is a ctypes array. The ctypes-types are bit a mystery to me, but the following seems to happen:

```python
> string = PLCTYPE_STRING
<class 'ctypes.c_char'>
> char_array = PLCTYPE_STRING * 80
<class '__main__.c_char_Array_80'>
> char_array.__type__
<class 'PyCArrayType'>
```

If are trying to read an array of e.g. floats, the type will be referred to as `PyCArrayType` internally.

stlehmann (approved):

stlehmann:
Nice, @RobertoRoos thanks for fixing this.

PR #237, opened 2021-04-09
Refactored testserver into multiple files

RobertoRoos:
I was going to tackle the notification stuff, but it bugs me the `testserver.py` file is over 1000 lines. So I suggest we split into multiple files (splitting the classes).

With the submodule `__init__.py` it is backwards compatible.


stlehmann:
Thanks for looking into this. I approve splitting the testserver in multiple modules. Any idea why the coverage went down?

RobertoRoos:
I'm glad.
No, I spent some time looking at it but I can't quite figure out the difference. None of the files were ignored for coverage?
This afternoon I might be able to add some tests to bring coverage back up I suppose.

RobertoRoos:
I don't think testserver.py was included in coverage to begin with: https://coveralls.io/github/stlehmann/pyads?branch=master
It's not one of the 9 files, I guess that's why coverage went down.

Do we want to omit the testserver code from coverage?

RobertoRoos:
This shouldn't have been rockets science, the test server files were excluded: https://github.com/stlehmann/pyads/blob/master/setup.cfg#L8 (but the mask now fails with the new files).

I'm in favour of _not_ omitting the test server files. I'll still try to increase coverage a bit (I promised that earlier but haven't had the opportunity yet).

RobertoRoos:
Checks are all green now.
I've added a few test methods but settled for a few `# pragma: no cover` comments - Mostly because quite a few statements are never going to be reached.

@stlehmann I'm ready for a merge when you are.

pyads/testserver/testserver.py
```diff
@@ -0,0 +1,331 @@
+# -*- coding: utf-8 -*-
```
stlehmann, line 1:
```suggestion
```

stlehmann: Guess this is a leftover from Python 2, we don't need this anymore

stlehmann (changes requested):
@RobertoRoos Thanks for your effort. Looks good to me. As for the `#pragma: no cover` statements: I don't think we should use them that extensively. If we decide to add testserver to coverage we should not pretend a better coverage by using them. I think it's better to remove the #pragmas and accept the lower coverage. It would be even better to add some test functions, though 😉.

stlehmann:
@RobertoRoos I made some modifications based on your work here and created a new PR #243. To avoid confusion I'll close this PR.

PR #238, opened 2021-04-10
added missing LINt type from datatype_map

chrisbeardy:


stlehmann:
Small but nice PR, thanks 👍

PR #243, opened 2021-05-12
Refactor testserver into multiple files

stlehmann:
This PR basically rebases #237 to master and fixes issues with coveralls.


stlehmann:
Thanks @RobertoRoos for this PR. I made some small changes regarding the whereabouts of the tests. I placed your additional tests for the AdvancedHandler in test_connection_class.py which is where also the tests for the BasicHandler live. I also removed the `# pragma` statements which eventually results in a lower coverage. But I don't mind that as long as we know where it comes from. Also we can improve coverage by time 😉

Issue #246, opened 2021-05-17
Add PLCTYPE_WSTRING support 


1130164523:
I have implemented the PLCTYPE_WSTRING variable related operations, can I merge it into the main branch? 
我已经实现了 PLCTYPE_WSTRING变量的相关操作，可以合并到主分支吗 ？

kryskool:
Hi @1130164523 

Can you  send a pull request 

Regards,

stlehmann:
This is solved with #274

PR #274, opened 2021-10-17
Add WSTRING support

stlehmann:
Addresses #246


stlehmann:
Looks like there needs to be some more effort here. Strings are handled special in read and write functions. So that needs to be applied to WSTRINGS as well.

stlehmann:
Now we also need some testing.

stlehmann:
I just made a Wireshark dump to see if the WSTRING is null-terminated with 1 or 2 Bytes: 

grafik

Clearly it is two bytes null-termination. So the testserver has to resemble this behaviour.

stlehmann:
I just have a use-case where I use WSTRINGs and discovered that adsSumRead and adsSumWrite also need specific WSTRING support

stlehmann:
Tested the implementation on my PC but couldn't get the test-cases pass, yet. Somehow the encoding is not right. Suggestions are welcome 😊

stlehmann:
@kryskool did you already do some testing on your setup with this branch?

stlehmann:
Todo: handle WSTRINGS right in adsSyncReadWriteReqEx2

PR #276, opened 2021-10-22
Test with Python 3.10

kryskool:
Python 3.10 is out

Be carefull, 3.10 need to be quote, instead of it recognize as 3.1

See #277


kryskool:
Hi @stlehmann 

Is my fix for Python 3.10 is enough 

I prefer to check attended values, before send the request (custom error message)

Regards,

stlehmann (approved):
Thanks for your PR. Looks great. I'll merge it right away 👍

Issue #228, opened 2021-03-10
Problem to communicate with PLC after setting up connection


mtroback:
Hi, This is probably not the best way to ask questions, maybe I should turn to stackoverflow instead, but I figure this is the best shot at getting good answers.

I've successfully set up pyads on my windows computer and I am able to use the pyads.read_by_name function without problems. It took me a while to figure out stuff, but it's working.

Now I have moved to my Raspberry Pi trying to communicate with the same PLC. This time it is trickier in the sense that I have to setup my own route. I have tried to follow the example in the Quickstart and I can successfully set up the route, but just for the sake of understanding from my point of view I have a quick question: The Client addresses are supposed to be the computer on which I am running python? So I choose the `CLIENT_NETID`  by using the computer IP and add .1.1? And the `CLIENT_IP` is the computers IP? The the `TARGET_*` is all the information of the PLC?

Then I don't understand why you use a NetId pointing at the computer localhost IP when you set up a connection. This was unsuccessfull for me, but it was successfull using a NetId of the target IP instead (tracebacks omitted for the sake of clarity, 192.168.3.1 is the raspberry pi, 192.168.3.9 is the PLC):
```
In [3]: pyads.add_route_to_plc('192.168.3.1.1.1', '192.168.3.1', '192.168.3.9', 'Administrator', '1', route_name='route')
Out[3]: True

In [4]: plc = pyads.Connection('127.0.0.1.1.1', 851)

In [5]: plc.open()
2021-03-10T14:35:01+0100 Error: Connect TCP socket failed with: 111
------
ADSError: ADSError: target port not found   ADS Server not started (6).

In [6]: plc = pyads.Connection('192.168.3.9.1.1', 851)

In [7]: plc.open()
2021-03-10T14:35:20+0100 Info: Connected to 192.168.3.9

In [8]: plc.read_device_info()
---------------------------------------------------------------------------
ADSError                                  Traceback (most recent call last)

ADSError: ADSError: target machine not found    Missing ADS routes (7).

In [9]: plc.read_by_name('MAIN.IN_1')
---------------------------------------------------------------------------
ADSError                                  Traceback (most recent call last)

ADSError: ADSError: target machine not found    Missing ADS routes (7).
```
So it looks like it is connected, but the route is missing? Can you spot what I'm doing wrong here?

I'm writing from my windows computer which has TwinCAT running and route set up in the background from the windows computer to the PLC I assume, could this be a problem? I am not connected from the windows computer.

[...]

PR #278, opened 2021-10-25
Improved documentation of how to set up routes, especially in Linux. …

mtroback:
…Changed example AMS netId to a more Beckhoff-alike one. As discussed in https://github.com/stlehmann/pyads/issues/228

I failed to execute Sphinx according to contribution guidelines, didn't manage to figure out what I did wrong (except maybe using windows)


doc/documentation/routing.rst
```diff
@@ -7,89 +7,55 @@ handled differently on Windows and Linux and is explained for both
...
-
-Get the AMS address of the local machine. This may need to be added to
-the routing table of the remote machine.
-
 .. note::
 
-  On Linux machines at least one route must be added before the
-  call to :py:func:`.get_local_address` will function properly.
-
-Optionally, a local AmsNetId can be manually set before adding a route.
```
stlehmann, line 50: We shouldn't skip this part. Setting local netid is necessary in some cases.

tests/test_connection_class.py
```diff
@@ -21,7 +21,7 @@
 # These are pretty arbitrary
 TEST_SERVER_AMS_NET_ID = "127.0.0.1.1.1"
 TEST_SERVER_IP_ADDRESS = "127.0.0.1"
-TEST_SERVER_AMS_PORT = pyads.PORT_SPS1
+TEST_SERVER_AMS_PORT = pyads.PORT_TC3PLC1
```
stlehmann, line 24: I don't know if this works. I think the testserver runs on PORT_SPS1 by default so also in the tests. I don't see a reason to change this.

stlehmann (changes requested):
Nice PR. Please have a look at the comments.

doc/documentation/routing.rst
mtroback: Interesting. Since I'm still a beginner of PLC programming I've had a few struggles recently. A small background: I have a python program that worked on Windows where I set up the Connection like this:
plc = pyads.Connection(self.target_ams_net_id, pyads.PORT_TC3PLC1)  # Works great on Windows, but not on Linux
To get it working in Linux I had to do add target ip:
plc = pyads.Connection(self.target_ams_net_id, pyads.PORT_TC3PLC1, self.target_ip)  # Works on both Linux and Windows
The difference is that on Linux I call the add_route_to_plc, which I don't on Windows.

But my question is probably: So there are situations where you do calls which require local netid but you have not set up a Connection?

stlehmann: Yes, if you want to change your local netid you need to do this before you create your Connection object.

grafik

mtroback:
I've updated the code according to comments, I guess you @stlehmann will get a notification from GitHub on this?

doc/documentation/connection.rst
```diff
@@ -24,8 +25,8 @@ A context notation (using ``with:``) can be used to open a connection:
 .. code:: python
 
    >>> import pyads
-   >>> plc = pyads.Connection('127.0.0.1.1.1', pyads.PORT_TC3PLC1)
-   >>> with plc:
+   >>> TARGET_NET_ID = '5.12.82.20.1.1'  # OR pyads.ads.adsGetNetIdForPLC('192.168.3.9') if unknown
```
stlehmann, line 28: Actually using 127.0.0.1.1.1 is on purpose as it refers to the locals AMS-ID of your local system. So I ask you to use this as a default. But bringing on the adsGetNetIdForPLC function is a nice touch.

doc/documentation/routing.rst
```diff
@@ -7,89 +7,69 @@ handled differently on Windows and Linux and is explained for both
...
-*target*. The *client* is your computer where pyads runs on. The
-*target* is you plc or remote computer which you want to connect to.
+*target*. The *client* is your computer where pyads runs. The
+*target* is your plc or remote computer which you want to connect to.
+
+For the communication to work there needs to be routes in both directions, both
+from the client to the target and from the target to the client. The route from
+the client to the target is created when a Connection is created, but the route
+from the target to the client needs to be created manually according to
+instructions below.
```
stlehmann, line 17: This only applies to Linux. On Windows one needs to add Routing over TwinCAT.

stlehmann (commented):
@mtroback thank you for your efforts. But speaking honestly I don't see that much improvement in your changes to the documentation. It is just written in a different way. I think I can't accept your PR but will take it as a reminder to further improve the docs on routing and also I might pick some of your suggestion when doing so. Sorry for this.

Anyway you also introduced some further PORT variables that haven't been included in the package. Would you like to create a separate PR just for these and I'm more then willing to merge these.

mtroback:
@stlehmann No problem! I've written this from a beginners perspective, reflecting the things I found difficult with no previous experience of ADS. This was a good training exercise for me both regarding ADS and github.

And in case any new beginners drops by they might get a hint of everything that is documented in this thread and might help them get started without disturbing you.

Once again, I'm very grateful for the work all of you have put in to get this working!

PR #290, opened 2021-11-25
moved write_list_by_name underneath write_by name

chrisbeardy:
just a tidy up, so all read_by_name methods are first then the write_by_names methods as opposed to jumbling them up


stlehmann:
We can do that. Brings more shape to the codebase. Don't know why the coverage decreased but can see no issues here. I'll merge it anyway.

Issue #291, opened 2021-11-25
Support WSTRING in structures.


stlehmann:
WSTRINGS are supported since version 3.3.9. Structure handling should also support this. This should mainly affect the functions dict_from_bytes and bytes_from_dict. Also some testing needs to be added.

stlehmann:
```python
if plc_datatype == PLCTYPE_STRING:
                    if str_len is None:
                        str_len = PLC_DEFAULT_STRING_SIZE
                    var_array.append(
                        bytearray(byte_list[index: (index + (str_len + 1))])
                        .partition(b"\0")[0]
                        .decode("utf-8")
                    )
                    index += str_len + 1
if plc_datatype == PLCTYPE_WSTRING:
    .... 
```

stlehmann:
Solved with #293

PR #293, opened 2021-11-29
Add WSTRING support for structures

stlehmann:
Targets #291


stlehmann:
This adds support for WSTRINGs in structures, both as fields and as arrays.
@chrisbeardy would you mind taking a look at this PR and giving a short review as this involves some changes on the codebase written by you.

pyads/ads.py
```diff
@@ -262,6 +263,11 @@ def size_of_structure(structure_def: StructureDef) -> int:
                 num_of_bytes += (str_len + 1) * size
             else:
                 num_of_bytes += (PLC_DEFAULT_STRING_SIZE + 1) * size
+        elif plc_datatype == PLCTYPE_WSTRING:
+            if str_len is not None:
+                num_of_bytes += (str_len + 1) * size
+            else:
+                num_of_bytes += (PLC_DEFAULT_STRING_SIZE + 1) * 2 * size
```
chrisbeardy, line 270: For W_STRINGs are these also terminated by a word so should it be + 2? It might be worth adding comments on these now explaining the plus one, and * 2 to explain the "magic" numbers.

Do we also need a * 2  and + 1/2 if the user specifies a length? If you declare a `WSTRING(10)` for example is that 20 + 1/2 bytes?

chrisbeardy: https://infosys.beckhoff.com/english.php?content=../content/1033/tc3_plc_intro/2529437323.html&id=

pyads/ads.py
```diff
@@ -319,6 +325,13 @@ def dict_from_bytes(
                         .decode("utf-8")
                     )
                     index += str_len + 1
+                elif plc_datatype == PLCTYPE_WSTRING:
+                    if str_len is None:
+                        str_len = 2 * PLC_DEFAULT_STRING_SIZE + 1  # use double default string size for WSTRING plus 1
+                    a = bytearray(byte_list[index: (index + (str_len + 2))])
```
chrisbeardy, line 331: similar comment to above, we have a mix of +1 and +2 here, line 330 vs line 331

tests/test_connection_class.py
```diff
@@ -1475,6 +1475,71 @@ def test_read_wstring(self):
...
+
+            # read struct with WSTRING array
+            val = self.plc.read_structure_by_name("wstring_array_struct", wstring_array_structure_def)
+            self.assertEqual({"name": ["hällo world", "foo bar"], "value": 10}, val)
+
+            # write struct with WSTRING array
+            self.plc.write_structure_by_name("wstring_array_struct", wstring_array_values, wstring_array_structure_def)
+            val = self.plc.read_structure_by_name("wstring_array_struct", wstring_array_structure_def)
+            self.assertEqual(wstring_array_values, val)
+
```
chrisbeardy, line 1543: we should probably add some more tests for dict_to_bytes and bytes_to_dict to cover uses of WSTRING

pyads/ads.py
stlehmann: > For W_STRINGs are these also terminated by a word so should it be + 2?
Agreed, should be +2 here. I think I missed that one.

> Do we also need a * 2 and + 1/2 if the user specifies a length? 
The length the user specifies with e.g. WSTING(10) defines the number of characters. For WSTRINGs a character takes two bytes. So the number of bytes needs to be 2* of this value. In the link you sent it says a WSTRING is terminated by a 0 but doesn't tell 1 or 2 Bytes, but by testing with my PLC I found out it is actually two Bytes.

stlehmann: Fixed it in #f9f7ffe

pyads/ads.py
stlehmann: Fixed it in f9f7ffe

tests/test_connection_class.py
stlehmann: Also we should add tests with fixed STRING/WSTRING sizes.

pyads/ads.py
chrisbeardy: If the null terminator is 2 bytes, should lines 268,  332, 421, 425, then in the tests lines 1501, 1512,  be `+ 2`, not `+ 1` ? 

Being picky, I'm not sure the `# if no str_len is given use default size` comments are required, I think this is fairly obvious from the `str_len = PLC_DEFAULT_STRING_SIZE`. The comments about the number of character explanations though are useful.

Other than that the code seems to fit with what it did previously.

stlehmann: > If the null terminator is 2 bytes, should lines 268, 332, 421, 425, then in the tests lines 1501, 1512, be + 2, not + 1 ?

All lines apply the schema `2 * (size + 1)`  so we have two Bytes null-terminator, the same in the tests. But as I applied the changes just yesterday in https://github.com/stlehmann/pyads/pull/293/commits/f9f7ffe054166b8dcd0d2affc15158e19dcc0c22 and https://github.com/stlehmann/pyads/pull/293/commits/6c606c38e92c38d2c9c2325d00ad7d867228a5f9 you might have missed them.

stlehmann: > Being picky, I'm not sure the # if no str_len is given use default size comments are required, I think this is fairly obvious from the str_len = PLC_DEFAULT_STRING_SIZE.

It' s a bit picky 😋 But you are basically right. I can't argue with that so I removed the comment.

chrisbeardy: > > If the null terminator is 2 bytes, should lines 268, 332, 421, 425, then in the tests lines 1501, 1512, be + 2, not + 1 ?
> 
> All lines apply the schema `2 * (size + 1)`  so we have two Bytes null-terminator, the same in the tests. But as I applied the changes just yesterday in https://github.com/stlehmann/pyads/pull/293/commits/f9f7ffe054166b8dcd0d2affc15158e19dcc0c22 and https://github.com/stlehmann/pyads/pull/293/commits/6c606c38e92c38d2c9c2325d00ad7d867228a5f9 you might have missed them.

I think I may have also been slightly stupid as `2 * (str_len + 1) ` is the same thing as ` (2 * str_len) + 2`...😬

PR #304, opened 2022-01-12
Added try-catch to notification close in symbol destructor

### LINKED ISSUE #303: Add try-catch to AdsSymbol destructor
(I'll put this issue here as the pyads-symbols package has not yet started.)

If an ADS connection is closed unexpectedly, the `AdsSymbol` destructor will throw errors when trying to close ads notifications. This could prevent other application destructors from being run.

I think it's best to quietly continue when the notification re-register fails here, since without an active connection nothing can be done in the first place.

RobertoRoos:
Fixes #303

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 * [#293](https://github.com/stlehmann/pyads/pull/2939) Support WSTRINGS in structures
 
 ### Changed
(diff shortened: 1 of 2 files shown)


stlehmann (changes requested):
Great work. Now we only need a Changelog entry for this change and I can merge it.

stlehmann (approved):
Nice work, thanks :)

stlehmann:
@RobertoRoos do you think you can find a way to add a test for this particular case=

PR #305, opened 2022-01-18
Support use of Beckhoff ADS library on win32 systems

cowo78:
Beckhoff ADS library can be built on win32 systems and used instead of TcAdsDll, thus simplifying install and usage procedures. See https://github.com/Beckhoff/ADS/pull/156 for more info.
Trying to load ADS library on win32 and disable router functions depending on undelying C library.

The above PR will *not* be merged as Beckhoff is not interested in keeping an usable C interface, so if you want to integrate the changes in your own fork let me know.


chrisbeardy:
I think I would like to stick as close as possible to the Beckhoff ADS.

PR #325, opened 2022-08-11
fix: add missing error code 29 and 30

kryskool:


chrisbeardy:
added alongside other changes. Sorry only just found the PR as could only find your repo at first from the issue link, otherwise would have merged this PR instead.

PR #369, opened 2024-02-08
Test for pull request #304 (Added try-catch to notification close in symbol destructor)

chhinze:
This is an extension to #304 (fixing #303) adding a test for the specific behavior. I have this problem repeatedly, when using `auto_update=true` for `AdsSymbols` and the server disconnects and, afterwards the `AdsSymbols` instances get garbage-collected.

I would love to see pull request #304 merged, so I added the requested test. The additional commit is directly added on top of the commit in the branch in #304. Without the try-catch in #304 the test fails. I hope the test is useful.

Is this what you had in mind @stlehmann?


RobertoRoos:
Stumbling across this again. Issue #303 is still relevant, thanks @chhinze for adding a test.

@stlehmann could we merge this?
I'm not sure if there are other maintainers too, @chrisbeardy maybe?

tests/test_testserver.py
```diff
@@ -48,6 +48,47 @@ def test_context(self):
...
+            # this raised an ADSError up to commit [a7af674](https://github.com/stlehmann/pyads/tree/a7af674b49b1c91966f2bac1f00f86273cbd9af8)
+            #  `clear_device_notifications()` failed, if not wrapped in try-catch as the server is no longer present.
+            test_int.__del__()
+        except pyads.ADSError as e:
+            raised_error = str(e)
+
+        # check no error was raised:
+        self.assertTrue(
+            not raised_error, f"Closing server connection raised: {raised_error}"
+        )
```
RobertoRoos, line 90: If you could be bothered, I find something like:

```python
test_server.stop()
time.sleep(0.1)  # Give server a moment

try:
    test_int.__del__()
except ExceptionType as err:
    self.fail(f"Closing server connection raised: {raised_error}")
```
more neat (see [SO](https://stackoverflow.com/questions/4319825/python-unittest-opposite-of-assertraises)).

RobertoRoos (approved):

chrisbeardy:
I have some plans this week or next to take a look at some of the Open PRs, as I have a few that I have "marked" in my own head as ready to merge so was going to spend some time doing it.

tests/test_testserver.py
chhinze: Sure. I would be happy to consider your suggestions. Just to be clear: do you mean by the comment that you find the comments too cluttering, or do you refer to the `self.assertTrue(...)` check, or both?

RobertoRoos: No, your in-code comments are fine. I mean the way of asserting an exception did _not_ get raised. I think the `except self.fail(...)` is cleaner than an `self.assert...` after the clause.

chhinze: Thank you for pointing it out and for the clarification. I changed it now with the recent commit [cbff3e8](https://github.com/stlehmann/pyads/pull/369/commits/cbff3e8975b80d2e478073478288c87851a52476).

tests/test_testserver.py
```diff
@@ -48,6 +48,42 @@ def test_context(self):
...
+        time.sleep(0.1)  # Give server a moment
+
+        try:
+            # some code, where test_int is cleared by the Garbage collector after the server was stopped
+            # (e.g. the machine with ADS Server disconnected)
+            # this raised an ADSError up to commit [a7af674](https://github.com/stlehmann/pyads/tree/a7af674b49b1c91966f2bac1f00f86273cbd9af8)
+            #  `clear_device_notifications()` failed, if not wrapped in try-catch as the server is no longer present.
+            test_int.__del__()
+        except pyads.ADSError as e:
+            self.fail(not raised_error, f"Closing server connection raised: {e}")
```
RobertoRoos, line 85: Thanks for updating. The syntax isn't correct though, `self.fail()` doesn't take in a verification parameter, only the string message:
```python
self.fail(f"Closing server connection raised: {e}")
```

PR #392, opened 2024-06-24
Fixes bug that causes AmsPorts in _adsDLL library to remain open

a-domino:
On Linux systems, if an exception was thrown from the adsAddRoute(...) function call in the Connection.open() function, after a port has been opened, there was no call to close that port. This lack of closing the port renders the underlying AmsPort to be unusable for later connections, and if all 128 AmsPorts end up in this state, all Connection objects are rendered useless.


chrisbeardy:
thanks for this, great spot, ill merge with some other changes soon

PR #389, opened 2024-06-07
Added TWINCATSDK environment variable to resolve Windows TcAdsll.dll

vincent-hve:
On Windows 11, for TwinCAT 4026.5, it seems TcAdsDll.dll has moved and we can not rely on TWINCAT3DIR. I have added to the resolution method to first search for a different variable TWINCATSDK, which seems to be the basis for the correct DLL. I've tested this on two separate systems.

PR #393, opened 2024-07-04
fix dll path issue with Twincat 4026

WeiTing1991:
I was facing some issues with the Twincat 4026 with fresh installation. try to fix the dll path (TcAdsDll.dll)
pyads
Fixed it with two dll_path both for 4024 ( earlier) and the new 4026 with pkg manager in windows platform. And tested in windows 10/11.


chrisbeardy:
thank you for this will look into merging this. related to #389

chrisbeardy:
closed by #389 , thank you @WeiTing1991 , unfortunately there was already a solution PR that had been submitted so I went with that solution. Both were valid. Thank you for your first contribution though and we would be happy to accept further PRs from you.

Issue #409, opened 2024-08-23
Incorrect type detection of string array in `parse_notification`


xiaoliu-heng:
the type could be wrong when use `auto_update` for c_char_Array like`STRING(28)`

<img width="604" alt="image" src="https://github.com/user-attachments/assets/63d28d35-30be-470a-ba42-c8f537d5809d">

chrisbeardy:
possibly related or same fix required for #356 and #405

RobertoRoos:
The fix in #410 is nice, but I think `AdsConnection` shouldn't re-implement the type parsing that already exists in `get_value_from_ctype_data()` already. I'm trying to squeeze it into PR #419 .

Issue #405, opened 2024-08-07
Issue reading arrays of strings


AliceJoby:
If there's an array of strings in the PLC, pyads will miss reading the extra byte needed for strings and won't be able to read past the first item in the array.
Ex. PLC has Array [0..10] of STRING(20)
pyads will parse the datatype from the PLC and determine the datatype to be pyads.symbol.c_char_Array_20_Array_10, which is correct. However, when determining the number of bytes needed to be read and reading the array, only 20 bytes will be read of the first array. Strings have an extra byte that need to be read. As a result, the last byte of the first item will become the first byte of the second item in the array, messing up the decoding.

If I put a breakpoint at https://github.com/stlehmann/pyads/blob/master/pyads/structs.py#L332 and modify the PLC datatype to be Array [0..10] of STRING(**21**), then the array will be read correctly.

chrisbeardy:
Thanks for this, there defo seems to be a bug here, effects both symbol reading and read_by_name, whether you pass type in or not, although if you do pass type in then you seem to only get back the first index of the array but decoded correctly, if you don't pass the type in you get back an array of c_char_Array_20 objects.

chrisbeardy:
looks linked / duplicate of #356 @AliceJoby there is a workaround in #356 if you need it while i fix the issue

chrisbeardy:
@RobertoRoos fancy taking a look at this? looks like #410 may have a simlar fix?

RobertoRoos:
Yeah, I'm having a look now.

I can reproduce this with the plc code:
```
VAR_GLOBAL
    // Sizes according to `SIZEOF()`:
    list_of_strings         : ARRAY[0..9] OF STRING;            // 810
END_VAR
```

and Python code:
```python
    with Connection(ams_net_id="127.0.0.1.1.1", ams_net_port=851) as plc:
        list_of_strings = plc.read_by_name("GVL_Example.list_of_strings")

    for item in list_of_strings:
        txt = item.value.decode()
        print(txt)
```

Issue #356, opened 2023-06-01
Problems with strings using read and readwrite


ChristianHemann:
Hello,

in a recent project we are using pyads to communicate with a plc, especially by using methods which we are accessing via readwrite. Also we need to read some long strings (length 10.000).
Both was working in pyads 3.3.4, but when updating to pyads 3.3.9 it starts failing. It seems as the problem is caused by the introduction of the helper function `type_is_string()` in `pyads_ex`

### First Problem: Reading long strings
In pyads 3.3.4 we were specifying the length of the string and pyads was returning the read bytes that we had to parse in the way we need them. 
Our code was looking like this:
```
value: bytes = plc_connection.read_by_name(
    "MAIN.VariableName",
    plc_datatype=pyads.PLCTYPE_STRING * (10000 + 1),
)
```

With pyads 3.3.9, pyads is using `type_is_string()`, which does not just check for `data_type == PLCTYPE_STRING`, but also checks for `"PyCArrayType"`. If it finds a string, which is now true for the datatype above, it is not using my string type with a custom length, but using the default string length of 1024 instead (`STRING_BUFFER * PLCTYPE_STRING`)

My suggestion to fix this would be to change 
`type_is_string(read_data_type)` into 
`type_is_string(read_data_type) and type(read_data_type).__name__ != "PyCArrayType"`
This would need to be done in `adsSyncReadReqEx2()` and `adsSyncReadWriteReqEx2()`. Maybe in other functions as well, which I have not tested. 

One change that I like in pyads 3.3.9 ist, that read_by_name is now returning a string, instead of a bytes-object.

### Second Problem: Calling Methods with a single string parameter
We have a method taking a single parameter of type string with a length of 255. It returns a struct, but that seems not to be the problem. 
When calling this functions via read_write, ADS is raising an Error, that the size is not correct. Our code looks similar to this:
```
handle_path = f"MAIN.MyObject#FunctionName"
handle = plc_connection.get_handle(handle_path)

parameter_type=pyads.PLCTYPE_STRING * (255 + 1)
parameters_value = "value".encode("ASCII")

return_value = plc_connection.read_write(
    pyads.constants.ADSIGRP_SYM_VALBYHND,
    handle,
    plc_read_datatype=return_type,
    plc_write_datatype=parameter_type,
    value=parameter_value,
)
```

In pyads 3.3.4 the length of the string was determined by `adsSyncReadWriteReqEx2` as:
```
write_data = write_data_type(*value)
[...]

PR #410, opened 2024-08-23
fix: use type_is_string to detect string type

xiaoliu-heng:
fix issue #409


chrisbeardy:
thanks for this, I've been meaning to look into this issue as i think it may be the same problem as #356 and #405. Since you're on the train of thought, could you check if it fixes these issues too (or the same solution is required), be good to close all 3 issues in one go!

xiaoliu-heng:
yes, I have already tested it on the actual PLC hardware.

xiaoliu-heng:
but it should only fixed the device notification

PR #419, opened 2024-09-13
String and string array fixes

RobertoRoos:
Trying to improve string handling to solve some of the open issues.

~~Frankly I'm a bit lost what the issues are exactly, I'm trying to go through them slowly.~~ EDIT: Okay, I think this is about it for now. I think this resolves all string issues?

Resolves:
- #420 
- #409 
- #405 

Should resolve (but I haven't reproduced the issue before):
- #356

Obsoletes other PRs:
- #410


pyads/pyads_ex.py
```diff
@@ -835,12 +835,9 @@ def adsSyncReadReqEx2(
...
 
-    if type_is_string(data_type):
-        data = (STRING_BUFFER * PLCTYPE_STRING)()
-    elif type_is_wstring(data_type):
-        data = (STRING_BUFFER * ctypes.c_uint8)()
-    else:
-        data = data_type()
+    # Strings were handled specifically before, but their sizes are contained and we
+    # can proceed as normal:
+    data = data_type()
```
RobertoRoos, line 840: Hm, I just realized this will break a usage for unknown string sizes, like:
```
plc.read_by_name("VAR", data_type=PLCTYPE_STRING)
```

RobertoRoos: Resolved in 0a259d8e , by assuming a singular character actually corresponds to an unknown size.

RobertoRoos:
Okay, I think I'm done. @chrisbeardy whadda think? @stlehmann maybe?

PR #412, opened 2024-08-29
Removed test from setup.py + Replaced all direct setup.py

RobertoRoos:
This should resolved #403 and is a first step towards #400

PR #426, opened 2024-09-23
Updated setuptools with `adslib` in wheel distribution and `pyproject.toml`

### LINKED ISSUE #425: Better distribution of `pyads` through platformed wheels
I was thinking about for a little bit and wanted to open issue this issue mostly to host some discussion.

Currently `pyads` is published only as a source package, no wheels at all: https://pypi.org/project/pyads/#files.
When installed on Linux/Mac, `adslib` is always compiled during install and the resulting `.so` library is dumped amidst the `pyads` source code, where it's dynamically loaded at runtime.

This definitely works, but I feel it's not according to modern standard and could be improved.

I would suggest we distribute mainly as wheels instead. So we'd publish:

- A wheel for Linux, already including the compiled `adslib` (and no `adslib` source)
- A wheel for MacOS, already including the compiled `adslib` (and no `adslib` source)
- A wheel for Windows, containing only the `pyads` source and no `adslib` source or binaries
- A source distribution for platforms falling through the cracks, with the `adslib` source and no binaries so they can be compiled during installation

This way most installations shouldn't require any compilation at all.  
I'm pretty far with an example of this already.

RobertoRoos:
I think this is pretty close to what I had in mind. This would resolved #425 .

Check the pipelines running on my fork to see what the distributions would look like.

And this PR will likely conflict with #412 and maybe with #399 .


RobertoRoos:
Hm, I just realized a pretty bug limitation: when installing in editable mode (`pip install -e .`) `adslib` isn't compiled and so the package is not usable. That needs to be fixed.

RobertoRoos:
Okay, I think I'm happy with this now.

PR #430, opened 2024-10-16
PLCTYPE_BYTE means uint8, so ADST_INT8 should mapped to PLCTYPE_SINT

lfncel:
PLCTYPE_BYTE means uint8, so ADST_INT8 should mapped to PLCTYPE_SINT


chrisbeardy:
you're not wrong, PLC_UBYTE also does not exist so not sure where this has come from, this should probably to PLCTYPE_USINT...I need to check what this impacts in the rest of the project too.

Issue #432, opened 2024-11-20
add_route_to_plc fails under Linux


vrees:
I want to connect to my Beckhoff PLC, CX5110, running TwinCAT 3.1, build 4024.62 via ADS protocol under Linux. While trying to **add_route_to_plc** an error occurred in the pyads lib:

![main_py](https://github.com/user-attachments/assets/be0ee47b-93c5-40d2-a28e-8fd1bc970250)

`onnected to pydev debugger (build 242.23339.11)
Traceback (most recent call last):
  File "/home/vrees/.local/share/JetBrains/IntelliJIdea2024.2/python-ce/helpers/pydev/pydevd.py", line 1570, in _exec
    pydev_imports.execfile(file, globals, locals)  # execute the script
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vrees/.local/share/JetBrains/IntelliJIdea2024.2/python-ce/helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "/home/vrees/work/git-repos/heatpump2influxdb/main.py", line 15, in <module>
    pyads.add_route_to_plc(CLIENT_NETID, CLIENT_IP, PLC_IP, PLC_USERNAME, PLC_PASSWORD,route_name=ROUTE_NAME)
  File "/home/vrees/work/git-repos/heatpump2influxdb/venv/lib/python3.11/site-packages/pyads/ads.py", line 200, in add_route_to_plc
    return adsAddRouteToPLC(
           ^^^^^^^^^^^^^^^^^
  File "/home/vrees/work/git-repos/heatpump2influxdb/venv/lib/python3.11/site-packages/pyads/pyads_ex.py", line 169, in wrapper
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/vrees/work/git-repos/heatpump2influxdb/venv/lib/python3.11/site-packages/pyads/pyads_ex.py", line 375, in adsAddRouteToPLC
    raise RuntimeError(f"Unexpected response from PLC: {data!r}")
RuntimeError: Unexpected response from PLC: b"\x03f\x14q\x00\x00\x00\x00\x06\x00\x00\x80\x05|\xca\xe0\x01\x01\x10'\x01\x00\x00\x00\x01\x00\x04\x00\x04\x07\x00\x00"
[...]

PR #437, opened 2024-12-02
adsGetNetIdForPLC: Solve too small buffer size for build 4026+

maxkofler:
Some PLCs that run TwinCAT >4026 send more bytes than expected when requesting the AMS NetID. This patch allocates more buffer size to the response to solve an error where the function would error out for no appearant reason.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 * [#384](https://github.com/stlehmann/pyads/pull/384) Enable processing of nested structures
 
+### Changed
+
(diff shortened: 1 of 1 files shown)


chrisbeardy:
Thanks for this, will look into merging shortly, do you think this will fix #432 ?

chrisbeardy:
Does this also work with no consequences for 4024?

maxkofler:
Thankyou for your fast response. I do not think that this would solve the other issue. This error originated in that the PLC sent more data than expected...

Regarding the tests, I can do some tomorrow. I will add another comment to let you know about that.

maxkofler:
@chrisbeardy I confirm that this change can be used with 4024. In fact, the buffer size can be way bigger even, so that's something we can think about doing to prevent such errors

PR #438, opened 2024-12-03
fixing read_list_by_name looping on MAX_ADS_SUB_COMMANDS case with lo…

SISAutomationIMA:
This pull request fixes an issue encountered when using the read_list_by_name function with structure_defs. If more than 500 tags are read and structure_defs is provided with a size matching the tags being requested, the for loop in sum_read raises an error. This happens because the tag list is sliced, but structure_defs is not.

To address this, I adjusted the for loop to ensure that the sliced chunk also includes the corresponding tags in structure_defs. This change resolves the issue.

### DIFF
--- a/pyads/connection.py
+++ b/pyads/connection.py
@@ -598,8 +598,9 @@ class Connection(object):
                                 list(structure_defs.keys()))  # type: ignore
 
             for data_name, structure_def in structure_defs.items():  # type: ignore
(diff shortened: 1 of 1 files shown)


chrisbeardy:
nice work, thank you, how do you fancy taking a look and seeing if this affects write_list_by_name too?

chrisbeardy:
Awesome thank you so much!

PR #469, opened 2025-06-27
Roberto roos feature/new setuptools

stlehmann:
This is the rebased version of PR #426


stlehmann:
@RobertoRoos great work. Everything seems to work as intended. Merging now.

Issue #433, opened 2024-11-25
Beckhoff Linux Distro: Connection closed by remote 


DanielJovanovski:
Hi,

I am running the Linux distribution provided by Beckhoff, which includes the TwinCAT runtime and the ADS router. I am trying to use the PyADS library to communicate with either the local runtime or another TwinCAT runtime on a Windows system.

In both cases, there is already an ADS route present. However, I am struggling to get PyADS to work as expected. After reviewing the PyADS documentation and troubleshooting, I suspect the issue might be related to how routes are handled on Linux, given that the Beckhoff distribution already has an integrated ADS router.

The PyADS documentation states:
_For Linux systems the route is created automatically on the client-side. For the target-side, you can use [add_route_to_plc()](https://pyads.readthedocs.io/en/latest/pyads.html#pyads.ads.add_route_to_plc)._

Since a route is already present, could this be the reason for my issues?

![image](https://github.com/user-attachments/assets/1f36749a-053d-43d6-b001-5c480fa532d3)

This is my Python code 
```
import pyads
import time

TARGET_NETID = "127.0.0.1.1.1"  

print(f"Attempting to connect to PLC at AMS Net ID: {TARGET_NETID}, Port: {pyads.PORT_TC3PLC1}")

try:
    with pyads.Connection(TARGET_NETID, pyads.PORT_TC3PLC1) as plc:
        print("Connection established successfully.")

        
        print("Checking PLC status...")
        state, device_state = plc.read_state()
        print(f"PLC State: {state}, Device State: {device_state}")

        if state != pyads.ADSSTATE_RUN:
            print("PLC is not in RUN state. Exiting...")
            exit(1)
        else:
            print("PLC is in RUN state. Proceeding to read variables...")

        while True:
            try:
                # Read the value of MAIN.bOutput_1
                print("Reading value of 'MAIN.bOutput_1'...")
                variable = plc.read_by_name('MAIN.bOutput_1', pyads.PLCTYPE_BOOL)
                print(f"Current value of 'MAIN.bOutput_1': {variable}")

                # Wait for 1 second before next read
                time.sleep(1)

            except pyads.ADSError as e:
                print(f"Error reading 'MAIN.bOutput_1': {e}")
                break

except pyads.ADSError as e:
    print(f"Failed to connect or communicate with PLC: {e}")
except KeyboardInterrupt:
    print("Script interrupted by user.")
finally:
    print("Exiting script.")
```

[...]

PR #479, opened 2025-12-12
Support for Beckhoff RT-Linux operating system (uses FreeBSD solution)

dewabe:
To support Beckhoff RT-Linux, we need to use a FreeBSD-based solution, even though the operating system itself is Linux-based.

My solution is to check whether TwinCAT is installed on Beckhoff RT-Linux (sudo apt install tc31-xar-um). If TwinCAT is installed - os.path.exists("/usr/bin/TcSystemServiceUm") - the system then uses a FreeBSD-based approach.

### DIFF
--- a/src/pyads/utils.py
+++ b/src/pyads/utils.py
@@ -8,16 +8,21 @@
 import functools
 import sys
 import warnings
-
+import os
 from .structs import SAdsSymbolEntry
 from .constants import PLC_STRING_TYPE_NAME, PLC_DEFAULT_STRING_SIZE
 
 from typing import Callable, Any, Optional
 
 
+def platform_is_beckhoff_rt_linux() -> bool:
+    """Workaround to identify Beckhoff RT-Linux as a FreeBSD-based solution instead of Linux."""
+    return os.path.exists("/usr/bin/TcSystemServiceUm")
+
+
 def platform_is_linux() -> bool:
     """Return True if current platform is Linux or Mac OS."""
-    return sys.platform.startswith("linux") or sys.platform.startswith("darwin")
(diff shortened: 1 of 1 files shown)


dewabe:
Yes, the issue appears when using Beckhoff RT-Linux, and that's the reason why I created additional function for that.

I assume Beckhoff RT-Linux shares some (most) of the features with TwinCAT/BSD, which leads that is_platform_linux() function doesn't fully work on this case - even though the system is Linux. One of the issues it that without the update, pyads cannot create connection to the system (and so, cannot read any variables).

The very same library (libTcAdsDll.so) is found in RT-Linux, like in FreeBSD (or TwinCAT/BSD) system.

chrisbeardy:
may fix #433

chrisbeardy:
I don't have access to a Beckhoff Linux IPC unfortunately. Out of interest, do you know what the command ` sys.platform` returns? I'm wondering if we use that instead if it returns anything useful to keep it in line with the other code.

chrisbeardy:
that makes sense and I suspect very much related to #433. The implementation of pyads never thought that Beckhoff may one day release a linux IPC so defo did not consider this in our implementation. It may also be fixed if we updated our ads dll we package too as that is quite old but that is a task for another day! I think this workaround will do for now. Can we just add a comment to the fucntion platform_is_linux_rt to say this is a workaround for when pyads is running on rt linux and say `see PR #479`. Just to make it clear as otherwsie I can see further down the line someone getting confused. Thanks.

dewabe:
Hi @chrisbeardy 
yes. And now I screwed my fork. And fixed it again. Commented it to be a workaround

Issue #457, opened 2025-05-14
Include NC error codes


SpareSimian:
Error codes for the NC subsystem can be found here:
https://download.beckhoff.com/download/document/automation/twincat3/TwinCAT_NC_Error_Codes_EN.pdf

PR #480, opened 2025-12-18
Added NC Error codes

dewabe:
Just a small thing, but enhancement anyway

Based on #457


dewabe:
To be honest, no. I did not do this manually. I started it manually, but no.. In my opinion, such a task is good to give to AI to pre-progress, and the validity can be checked afterwards. Obviously, not all the lines are double checked by me but many are.

Anyway:
Typing is removed now!

Issue #488, opened 2026-01-14
Fix warnings in `pytest` under Python 3.14


RobertoRoos:
Following 
- #486 

See output under: https://github.com/stlehmann/pyads/pull/486#issuecomment-3748726477

PR #490, opened 2026-01-14
Replaced naive `datetime`s with timezoned ones

RobertoRoos:
Together with #489 , this contributes to
- #488


RobertoRoos:
Coverage is decreased because a custom class has been removed. I suggest to merge anyway.  
@chrisbeardy FYI

RobertoRoos:
Oh yeah, this should already be in the changelog.

PR #494, opened 2026-01-14
Removed pytest warning from deleting from callback_store

### LINKED ISSUE #493: `pytest.PytestUnraisableExceptionWarning` for a `KeyError` in tests
Following
- #488 

See for example: https://github.com/stlehmann/pyads/actions/runs/20993886363/job/60345793297#step:7:135

It's only in `test_testserver.py::test_server_disconnect_then_del_device_notification`

When a debugger is attached, the issue cannot be shown.

RobertoRoos:
Resolves #493:
- #493

### DIFF
--- a/src/pyads/pyads_ex.py
+++ b/src/pyads/pyads_ex.py
@@ -1371,12 +1371,7 @@ def adsSyncDelDeviceNotificationReqEx(
     pAmsAddr = ctypes.pointer(adr.amsAddrStruct())
     nHNotification = ctypes.c_ulong(notification_handle)
     err_code = adsSyncDelDeviceNotificationReqFct(port, pAmsAddr, nHNotification)
-
-    store_key = (adr, notification_handle)
-    callback_store.pop(store_key, None)
-    # It is not correct if this notification is not in the callback store, but we can't
-    # do anything about it, so avoid exceptions
-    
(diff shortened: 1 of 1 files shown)


src/pyads/pyads_ex.py
```diff
@@ -1371,7 +1371,12 @@ def adsSyncDelDeviceNotificationReqEx(
     pAmsAddr = ctypes.pointer(adr.amsAddrStruct())
     nHNotification = ctypes.c_ulong(notification_handle)
     err_code = adsSyncDelDeviceNotificationReqFct(port, pAmsAddr, nHNotification)
-    del callback_store[(adr, notification_handle)]
+
+    store_key = (adr, notification_handle)
+    callback_store.pop(store_key, None)
+    # It is not correct if this notification is not in the callback store, but we can't
+    # do anything about it, so avoid exceptions
```
RobertoRoos: @stlehmann what about a literal `warning` then:

```suggestion
    store_key = (adr, notification_handle)
    try:
        callback_store.pop(store_key)
    except KeyError:
        warning.warn(f"Tried to de-register an unknown notification {store_key}", category=warning.RuntimeWarning)
    # It is not correct if this notification is not in the callback store, but we can't
    # do anything about it, so avoid exceptions
```

Issue #487, opened 2026-01-14
Including binary wheel for Windows


RobertoRoos:
Following #425 , maybe we should also include a binary wheel for Windows.

Issue #483, opened 2026-01-13
GLIBCXX_3.4.32 not found


aidiaz:
Hello!

Latest release breaks our current development setup. We have been using pyads for a year and a half an never had issues but with the current version we faced the following error. 
```
manager  |   File "/usr/local/lib/python3.12/site-packages/pyads/__init__.py", line 10, in <module>
manager  |     from .ads import (
manager  |   File "/usr/local/lib/python3.12/site-packages/pyads/ads.py", line 50, in <module>
manager  |     from .pyads_ex import (
manager  |   File "/usr/local/lib/python3.12/site-packages/pyads/pyads_ex.py", line 96, in <module>
manager  |     _adsDLL = ctypes.CDLL(adslib_path)
manager  |               ^^^^^^^^^^^^^^^^^^^^^^^^
manager  |   File "/usr/local/lib/python3.12/ctypes/__init__.py", line 379, in __init__
manager  |     self._handle = _dlopen(self._name, mode)
manager  |                    ^^^^^^^^^^^^^^^^^^^^^^^^^
manager  | OSError: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.32' not found (required by /usr/local/lib/python3.12/site-packages/adslib.so)

```

Going back to version 3.5.0 shows no errors.

aidiaz:
Thanks for the fast reply! I was quite unsure about how to start

We use docker to build and run python:3.12.1-slim based container, and it's my only guess about the .so building difference. Tomorrow I will try installing the missing packages on our dockerfile to run 3.5.1

stlehmann:
The error indicates that the version of your libstdc++.so.6 doesn't match the required version. Could you try updating with 

```
apt-get update && apt-get install -y libstdc++6
```

chrisbeardy:
@dewabe can you confirm version 3.5.1 works on the offical Beckhoff linux version if you still have access

RobertoRoos:
@aidiaz , your short-term workaround will be including `--no-binary=pyads` in the `pip install` I think.

@stlehmann @chrisbeardy shall we consider moving to `cibuildwheel` then? I think it will add to our future-proofness.

stlehmann:
@RobertoRoos thanks for digging into this. Does that mean your Pipeline builds with the older version of *libstdc++.so.6*? So maybe *auditwheel* resolves this in the correct way. Looks good to me. As far as I can see only the Upload is not yet working, right?

PR #496, opened 2026-01-15
Changed to `cibuildwheel` for wheel building across platforms

RobertoRoos:
Resolves:
- #487 
- #483 

I think this pipeline is now according what's considered modern. 

**Important!** Now PyPi authentication is done based on a trusted publisher. This is the recommended authentication for deploy jobs like these. I set it up for my own testpypi and it's easy. See:
https://docs.pypi.org/trusted-publishers/using-a-publisher/ 

The test-version of these can be seen on my fork: https://github.com/RobertoRoos/pyads/pull/1
Release results can be seen on my Test-PyPi: https://test.pypi.org/project/pyads-Roberto/#history
You could even try a test install with: `pip install pyads_roberto -v --index-url https://test.pypi.org/simple/`

This will obviously need a squash-merge... :')

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13', '3.14']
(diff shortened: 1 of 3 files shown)


stlehmann:
@RobertoRoos ist this MR ready to merge ... What do you think?

Issue #495, opened 2026-01-15
Warnings from `setuptools`


RobertoRoos:
During package build we get warnings in the output:
```
...

Successfully installed setuptools-80.9.0 wheel-0.45.1
  WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager, possibly rendering your system unusable. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv. Use the --root-user-action option if you know what you are doing and want to suppress this warning.
  Installing build dependencies ... done
  Running command Getting requirements to build wheel
  /tmp/pip-build-env-u6u6bs0w/overlay/lib/python3.14/site-packages/setuptools/config/_apply_pyprojecttoml.py:82: SetuptoolsDeprecationWarning: `project.license` as a TOML table is deprecated
  !!

          ********************************************************************************
          Please use a simple string containing a SPDX expression for `project.license`. You can also use `project.license-files`. (Both options available on setuptools>=77.0.0).

          By 2026-Feb-18, you need to update your project and remove deprecated calls
          or your builds will no longer be supported.

          See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
          ********************************************************************************

  !!
    corresp(dist, value, root_dir)
  /tmp/pip-build-env-u6u6bs0w/overlay/lib/python3.14/site-packages/setuptools/config/_apply_pyprojecttoml.py:61: SetuptoolsDeprecationWarning: License classifiers are deprecated.
  !!

          ********************************************************************************
          Please consider removing the following classifiers in favor of a SPDX license expression:

          License :: OSI Approved :: MIT License

          See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
          ********************************************************************************

  !!
    dist._finalize_license_expression()
  /tmp/pip-build-env-u6u6bs0w/overlay/lib/python3.14/site-packages/setuptools/dist.py:759: SetuptoolsDeprecationWarning: License classifiers are deprecated.
  !!

          ********************************************************************************
          Please consider removing the following classifiers in favor of a SPDX license expression:

          License :: OSI Approved :: MIT License

          See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
          ********************************************************************************

  !!
...

PR #497, opened 2026-01-15
Minor update to pyproject.toml for license fields

RobertoRoos:
Resolves
- #495

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## Unreleased
+
+### Changed
+* Minimum Python version changed to 3.9 (Python 3.8 is no longer supported)
+* [#495](https://github.com/stlehmann/pyads/issues/495) Organizing of metadata in `pyproject.toml` 
+
 ## 3.5.1
 
 ### Added


chrisbeardy:
It feels a bit wrong dropping support for a python version over a purely static bit of metadata telling people the licence, i know its EOL but i bet a lot of people using pyads may still be on 3.8. I personally have no projects still on 3.8 but i do on 3.9 and I know the automation industry moves slow. Is this the only way of telling people license terms these days "included" with the package, because its in the repo. What does this metadata do, just make it appear on pypi as MIT? I'd be more inclined to drop python version support because of a new feature for example.

chrisbeardy:
Having said that, I do see a very high value and do want to move to a more modern build platform including pyproject.toml As I definitely think that will help maintain things in the future and also stuff like the what we're doing with the wheel builds will definitely help across multi platforms and probably just generally Making things more maintainable. So ultimately if keeping 3.8 support means we can't modernize our build platform then I would also be happy to drop support for that version.

RobertoRoos:
> It feels a bit wrong dropping support for a python version over a purely static bit of metadata telling people the licence

I won't miss 3.8, but this is indeed a peculiar point for the `setuptools` people to drop the axe for.

Actually, I was already curious so I tried to keep up 3.8 and came up with this:
https://github.com/RobertoRoos/pyads/tree/bugfix/toml-license-3.8

By using `dynamic` the issue can be bypassed it seems. It's just that I can't really test the results of this change, I don't know how the info is used anyway.
Personally I say we don't bend over backwards and just drop it.

chrisbeardy:
This made me chuckle, just compiled some software for my PC and got this warning :
```
writing top-level names to pyrsistent.egg-info/top_level.txt
writing manifest file 'pyrsistent.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
adding license file 'LICENSE.mit'
writing manifest file 'pyrsistent.egg-info/SOURCES.txt'
* Building wheel...
/usr/lib/python3.13/site-packages/setuptools_scm/_integration/deprecation.py:7: UserWarning: setup.cfg: at [metadata]
version = attr: ... is forcing setuptools to override the version setuptools-scm did already set
When using setuptools-scm it's invalid to use setuptools dynamic version as well, please remove it.
Setuptools-scm is responsible for setting the version, forcing setuptools to override creates errors.
  warnings.warn(
[01/18/26 19:30:37] WARNING  toml section missing       pyproject_reading.py:215
                             PosixPath('pyproject.toml'                         
                             ) does not contain a                               
                             tool.setuptools_scm                                
                             section                                            
/usr/lib/python3.13/site-packages/setuptools/dist.py:759: SetuptoolsDeprecationWarning: License classifiers are deprecated.
!!

        ********************************************************************************
        Please consider removing the following classifiers in favor of a SPDX license expression:

        License :: OSI Approved :: MIT License

        See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
        ********************************************************************************

!!
  self._finalize_license_expression()
running bdist_wheel
running build
running build_py
```

We're not the only ones

stlehmann:
So I recon we can go ahead and merge this. @RobertoRoos would you mind adding an entry to the changelog which includes the info that Python 3.8 is no longer supported?

PR #508, opened 2026-04-04
Use upstream adslib directly from Beckhoff

### LINKED ISSUE #507: Implement adslib directly from Beckhoff repository

stlehmann:
Resolves #507


RobertoRoos:
It would be easier for us to rely on the upstream directly.

The downside is it will require more dependencies on the client system, in case we don't have a pre-compiled wheel for them. @stlehmann do you think this is acceptable? I myself have no real idea how complete our wheel coverage is right now.

EDIT: Also relevant for editable installs from source.

stlehmann:
@RobertoRoos I'm in contact with Patrick Bruenn from Beckhoff discussing how to make adslib directly compilable to a shared library. So far it looks promising.

RobertoRoos:
@stlehmann what just happened in this repo? All PRs got closed, was that you? Those PRs (and this one too) suddenly show 900+ commits diff. Which is about the entire size of this repo.

stlehmann:
@RobertoRoos sadly I must admit that this was me, indeed. Yesterday I accidentally force-pushed another repository to pyads. I tackled by force-pushing the current version of pyads again. Unfortunately all PRs got closed. I haven't found a way to deal with this, yet. Any help is welcome. I'm sorry for the mess 😒.

RobertoRoos:
Oof 😅 
That's alright, these things happen.

I fetched locally without pulling and it doesn't show any differences with the upstream, so all commit references, tags, etc. should still work.

We should re-open the closed PRs, but maybe put them in 'Draft' because I think they would require a touchup before they could be merged.

stlehmann:
Tell me ... 😮‍💨
Now I set up a branch protection rule to prevent this from happening again, hopefully.

Yes, I would like to re-open the PRs again. Don't know really how to do this because they refuse to enable the "REOPEN" button:

grafik

If someone has a good idea ...

chrisbeardy:
Oh dear, I thought something was up. I feel like we must be able to recover from this somehow, did it rewrite history? If not we may be able to rewrite it back.

RobertoRoos:
@chrisbeardy , the history is still identical, I checked that yesterday. See my comment above.

I was experimenting with trying to re-open #419 . I just rebased and force-pushed that branch, but the existing PR still shows this `901 commits` diff.

This seems to be a known situation, see for example: https://gist.github.com/robertpainsi/2c42c15f1ce6dab03a0675348edd4e2c 

EDIT: @stlehmann are you allowed to re-open #419 now? I reset my branch back to the last commit of `master` before the master-force-push, like the linked snippet suggests.

stlehmann:
Conversation continues in #511

Issue #420, opened 2024-09-13
Strings are always read a fixed length


RobertoRoos:
Another one for the pile of string related issues: it looks like strings are read at a fixed 1024 characters in e.g. `plc.read_by_name()`:

```python
# pyads_ex.py:
    if type_is_string(data_type):
        data = (STRING_BUFFER * PLCTYPE_STRING)()
```
```python
# constants.py:
STRING_BUFFER: int = 1024
```

Which seems unnecessary because the returned type of a string looks like `"STRING(80)"` and would contain the right size perfectly fine.

EDIT: Same for `WSTRING` (which is double-byte USC-2 encoded)

PR #513, opened 2026-04-29
String and string array fixes

RobertoRoos:
Recreation of faulty PR:
- #419 

Trying to improve string handling to solve some of the open issues.

~~Frankly I'm a bit lost what the issues are exactly, I'm trying to go through them slowly.~~ EDIT: Okay, I think this is about it for now. I think this resolves all string issues?

Resolves:
- #420 
- #409 
- #405 

Should resolve (but I haven't reproduced the issue before):
- #356

Obsoletes other PRs:
- #410

### DIFF
--- a/src/pyads/constants.py
+++ b/src/pyads/constants.py
@@ -31,10 +31,6 @@ PLC_DEFAULT_STRING_SIZE: int = 80
 MAX_ADS_SUB_COMMANDS: int = 500
 
 
-class PLCTYPE_WSTRING:
-    """Special dummy class for handling WSTRING."""
-
-
 # plc data types:
 PLCTYPE_BOOL = c_bool
 PLCTYPE_BYTE = c_ubyte
@@ -45,6 +41,7 @@ PLCTYPE_LREAL = c_double
 PLCTYPE_REAL = c_float
 PLCTYPE_SINT = c_int8
 PLCTYPE_STRING = c_char
+PLCTYPE_WSTRING = c_wchar
 PLCTYPE_TOD = c_int32
 PLCTYPE_UBYTE = c_ubyte
 PLCTYPE_UDINT = c_uint32
--- a/tests/test_connection_class.py
+++ b/tests/test_connection_class.py
@@ -232,11 +232,9 @@ class AdsConnectionClassTestCase(unittest.TestCase):
         # Assert that the server received the correct command
         self.assert_command_id(requests[0], constants.ADSCOMMAND_READ)
 
-        # The string buffer is 1024 bytes long, this will be filled with \x0F
-        # and null terminated with \x00 by our test server. The \x00 will get
-        # chopped off during parsing to python string type
-        expected_result = "\x0F" * 1023
-        self.assertEqual(result, expected_result)
+        # We are reading only a single character, which the test-server defaults at 0
+        expected_result = "\x00"
+        self.assertEqual(expected_result, result)
 
     def test_write_uint(self):
         value = 100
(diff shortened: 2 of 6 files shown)


stlehmann (commented):
In process...

tests/test_connection_class.py
```diff
@@ -1452,7 +1452,7 @@ def test_read_wstring(self):
         var = PLCVariable(
             "wstr",
             expected1.encode("utf-16-le") + b"\x00\x00",
-            constants.ADST_WSTRING, f"WSTRING({len(expected1)})"
+            constants.ADST_WSTRING, "WSTRING(80)"
```
stlehmann, line 1455: This is what's causing the CI failure. The fixture now declares `"WSTRING(80)"`, but the `value` bytes are still only as long as the actual content (`len(expected1) * 2 + 2 = 26 bytes`). On a real PLC a `WSTRING(80)` variable always occupies the full declared size (`80 * 2 + 2 = 162 bytes`), padded with zeros after the null terminator.

The `adsSumRead` path infers scalar-vs-array by dividing the symbol's reported `.size` by the per-element byte length derived from the declared type — so when `.size` (26) doesn't match `element_size` (162), it concludes it must be an array of `int(26/162) == 0` elements and returns `[]`.

Padding the value to the declared byte length fixes the immediate failure and makes the fixture match real PLC behaviour:

```suggestion
            (expected1.encode("utf-16-le") + b"\x00\x00").ljust(80 * 2 + 2, b"\x00"),
            constants.ADST_WSTRING, "WSTRING(80)"
```

---
_Generated by [Claude Code](https://claude.ai/code)_

src/pyads/constants.py
```diff
@@ -45,6 +41,7 @@ class PLCTYPE_WSTRING:
 PLCTYPE_REAL = c_float
 PLCTYPE_SINT = c_int8
 PLCTYPE_STRING = c_char
+PLCTYPE_WSTRING = c_wchar
```
stlehmann, line 44: Small platform note: `c_wchar` is 2 bytes on Windows but 4 bytes on Linux/macOS. The good news is that `bytes(ctypes_array)` always returns raw memory regardless of ctypes type interpretation, so `bytes(data).decode("utf-16-le").rstrip("\x00")` in `get_value_from_ctype_data` works correctly on both platforms.

One side-effect though is that the read buffer allocated in `adsSyncReadReqEx2` becomes `c_wchar * STRING_BUFFER = 4 * 1024 = 4096 bytes` on Linux, where the old code used `c_uint8 * STRING_BUFFER = 1024 bytes`. Requesting more bytes than the symbol holds should be harmless in practice (the ADS library returns what's available and the length check is skipped for strings), but it's worth a quick sanity check if you have a live PLC to test against.

---
_Generated by [Claude Code](https://claude.ai/code)_

stlehmann (commented):
Thanks for putting this together — the overall direction is really good. Centralising string/WSTRING decoding through `get_value_from_ctype_data`, removing the old `PLCTYPE_WSTRING` sentinel class in favour of a real ctypes type, and fixing the `STRING(n)` null-terminator off-by-one in `get_type_from_str` are all solid improvements.

Left two inline notes — one is a concrete suggestion to fix the failing `test_read_wstring` CI test, the other is just a platform portability observation. Otherwise the change looks good to me; once you rebase onto current `master` and push, CI should give a fresh green run.

---
_Generated by [Claude Code](https://claude.ai/code)_

RobertoRoos:
I'll try to have another look at this in the coming week. But it feels like a decade since I worked on this, so it's not so easy 😅

PR #515, opened 2026-04-30
allow buffer length of zero for Adswrite / adsSyncWriteReqEx

### LINKED ISSUE #335: Sending write command without variable value (length 0)
I am trying to send an `AdsSyncWriteReqEx` command without any data using `pyads`

The equivalent C API command I am trying to send is using a `nullptr` as data (with length 0)
```c
AdsSyncWriteReqEx(nPort, &mAddr, 0xF088, 0x00000000 | (3 & 0x000000FF), 0, nullptr);
```
and is taken from the [TwinCAT 3 | Corrected timestamps - ADS consumer](https://infosys.beckhoff.com/english.php?content=../content/1033/corrected_timestamps/6313305611.html&id=7695130188588900136) example.

I initially tried `plc.write()` which didn't work and also tried using the ADS bindings in `pyads` more directly after opening the ADS connection with `plc.open()`

```python
from pyads.pyads_ex import adsSyncWriteReqEx

adsSyncWriteReqEx(plc._port, plc._adr, 0xF088, 0x00000000 | (3 & 0x000000FF), None, None)
```

but I run into the following error which I think indicates that `None` is not a valid replacement for `nullptr`.

```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In [5], line 2
      1 from pyads.pyads_ex import adsSyncWriteReqEx
----> 2 adsSyncWriteReqEx(plc._port, plc._adr, 0xF088, 0x00000000 | (3 & 0x000000FF), None, None)

File pyads/pyads_ex.py:495, in adsSyncWriteReqEx(port, address, index_group, index_offset, value, plc_data_type)
    493     data = value
    494 else:
--> 495     data = plc_data_type(value)
    497 data_pointer = ctypes.pointer(data)
    498 data_length = ctypes.sizeof(data)

TypeError: 'NoneType' object is not callable
```

Is there any way I could send this command using `pyads`? How to work with `nullptr` in python?
I am thankful for any suggestions

CagtayFabry:
This small change allows `adsSyncWriteReqEx` to handle write requests with a buffer length of zero (no data)

This edge case is necessary to support ADS timestamp correction based on time providers configured in TwinCAT 3.

This feature requires a recent version of `adslib` to be installed (implemented upstream since commit [standalone/AdsLib: allow AdsWrite for zero length buffer](https://github.com/Beckhoff/ADS/commit/8bee2f64b62724aa5313997d8e60d0c68c789da6) )

closes #335

ref: https://github.com/Beckhoff/ADS/issues/264


CagtayFabry:
What are your thoughts on merging this soon and including it in the next release @stlehmann @chrisbeardy ?

Anything that is still missing here?
I am not sure how to add an appropriate test but I could give it a try.

chrisbeardy:
I think this looks good, I'll sort out merge conflict then merge it. I think we updated the adslib so should be ok

CagtayFabry:
The merge conflict was only the added changelog line, I quickly resolved it. Should be ready now 🚀

stlehmann:
Sorry for my late part taking. Thanks @CagtayFabry for this issue and your PR on this. As I'm not so familiar with it: Could you give a short use-case when to use ADS timestamp correction? This would be something that could go to the docs I suppose.

I also have two suggestions two suggestions / comments to the PR:
- We should include "None" in the type annotation of the function.
- Also adding a test for this use-case seems sensible

@chrisbeardy what do you think?

Kind Regards
Stefan

chrisbeardy:
These are probably wise additions, more so the None type addition is probably warranted and overlooked by me on review. Documenting the use case probably not so much but could be useful for anyone wondering why they can send zero length data. I just saw this as a small change bringing us in line with the Beckhoff ADS repos changes.

CagtayFabry:
> Sorry for my late part taking. Thanks @CagtayFabry for this issue and your PR on this. As I'm not so familiar with it: Could you give a short use-case when to use ADS timestamp correction? This would be something that could go to the docs I suppose.
> 
> I also have two suggestions two suggestions / comments to the PR:
> 
> * We should include "None" in the type annotation of the function.
> * Also adding a test for this use-case seems sensible
> 
> @chrisbeardy what do you think?
> 
> Kind Regards Stefan

Hi @stlehmann ,

sorry I didn't include more explanation in the PR.

For a general reference I think it is good to look into the TwinCAT documentation: https://infosys.beckhoff.com/content/1033/tc3_grundlagen/6311995659.html?id=3819454857031399157

In an nutshell, as you know the PLC operates it's own "real-time" clock that is the basis of it's operation. When starting the TwinCAT XAR this clock is initially sinced to the current system time. Hopwever, over time - especially when the real-time system is running for multiple weeks or month uninterrupted - this internal clock can shift continously and as a result deviate considerably from the "true" external time (e.g. what we get from NTP servers). Due to the way the real-time system is operating, there is no way of adjusting the internal real-time clock.

When sending ADS requests, by default the associated returned timestamps are given in the real-time clock timeframe.
In my use case I had to synchronize these readings with events that are outside of the real-time context (e.g. other IoT devices, external software, file systems, surveillance cameras etc.)

With the TwinCAT corrected tiemstamps features, it is possible to add external reference clocks (e.g. NTP providers) that are then monitored in TwinCAT. TwinCAT continously calculates the current offset between its internal real-time clock and the external time-provider.

When using the ADS configuration (which requires this PR), one can configure which timestamps are sent back for a given PLC connection.
Instead of the usual *real-time timestamp* one can switch over to having all timestamps be *real-time timestamp* + *offset(provider)*, which makes them align and comparable with the external NTP time server or other reference clocks.

I have implemented the basic commands from the documentation for pyads, happy to provide this in a follow up PR if you are interested.

PR #519, opened 2026-05-14
Fix: map ADST_INT8 to PLCTYPE_SINT instead of PLCTYPE_BYTE

AskicV:
## Problem
ADST_INT8 represents TwinCAT SINT, a signed 8-bit integer (-128 to 127).
It was incorrectly mapped to PLCTYPE_BYTE (c_ubyte, unsigned 0-255),
causing an exception when writing values to variables of this data type.

## Root cause
In src/pyads/constants.py:
ADST_INT8: PLCTYPE_BYTE  ← c_ubyte is unsigned (0-255)

## Fix
ADST_INT8: PLCTYPE_SINT  ← c_int8 is signed (-128 to 127)


stlehmann:
@AskicV thanks for this....merged it.

PR #525, opened 2026-05-31
Allow None in write type annotations for zero-length write requests

chrisbeardy:
Human Note: AI generated but seems to check out

PR #515 added support for zero-length write requests (plc_data_type=None) in adsSyncWriteReqEx but did not update the type annotations to reflect that None is now an accepted value. Update the annotations on adsSyncWriteReqEx and Connection.write to Optional[Type], and add a test covering the zero-length write use-case.

https://claude.ai/code/session_01QKx2yv6cnjKgJAeHkU6JEG

### DIFF
--- a/src/pyads/connection.py
+++ b/src/pyads/connection.py
@@ -282,7 +282,7 @@ class Connection(object):
 
     def write(
             self, index_group: int, index_offset: int, value: Any,
-            plc_datatype: Type["PLCDataType"]
+            plc_datatype: Optional[Type["PLCDataType"]]
     ) -> None:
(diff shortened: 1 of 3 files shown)


stlehmann (approved):
Nice and clean PR. @chrisbeardy which agent did you use for this?

chrisbeardy:
> Nice and clean PR. @chrisbeardy which agent did you use for this?

All done from the comfort of my sofa using Claude code from the phone app. Remarkable. I use Claude Code extensively at work now and given the right amount of oversight its a powerful tool. I should probably let it have a go more on pyads, especially on existing problems and review.

PR #531, opened 2026-06-13
fix: rename adslib.so -> AdsLib.so to match upstream library name

pbruenn:
The upstream AdsLib project renamed the library output from libadslib.so
to libAdsLib.so in https://github.com/Beckhoff/ADS/commit/ca978d769b3f75fa2f6b8239d8be578bcf116045.

Update all references in documentation, setup.py, and pyads_ex.py to
use the correct case-sensitive library name.

Also bump the adslib submodule to 6788651.

Signed-off-by: Patrick Bruenn <p.bruenn@beckhoff.com>
Assisted-by: OpenCode:claude-opus-4.6

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ pyads - Python package
 [![Downloads](https://pepy.tech/badge/pyads/week)](https://pepy.tech/project/pyads)
 
 This is a python wrapper for TwinCATs ADS library. It provides python functions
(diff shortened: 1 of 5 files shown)


stlehmann (approved):
@pbruenn Thanks for this update

stlehmann:
I included these changes in #537.

PR #539, opened 2026-06-21
testserver: unique notification handles + survive bad input

yannS2016:
## testserver: unique notification handles + survive bad input

Three small robustness fixes to the `AdvancedHandler` test server, found while using it to exercise an EPICS ADS driver's bulk-read and notification paths against the full symbol set of a real PLC project.

### 1. Unique notification handles across variables
`PLCVariable.register_notification` handed out colliding handles. `notification_count` is a class attribute, but `self.notification_count += 1` rebinds it per-instance, so the **first** notification registered on every variable returns the same handle (`10`), the second `11`, and so on. A client subscribed to more than one variable then cannot tell the notifications apart. Increment the shared class counter so handles are globally unique, as real TwinCAT returns them.

### 2. Survive an abrupt client disconnect
`AdsClientConnection.run` called `recvfrom()` with no guard. A client killed mid-request drops the connection abruptly, raising `ConnectionResetError` and crashing the handler thread, which leaves the server unable to serve the next connection. Treat the reset as a disconnect.

### 3. Survive a handler exception
A command handler that raised killed the connection's handler thread, leaving the client's later requests unanswered until they time out. `handle_read_write` does `write_data.decode()` on the read-write payload, which raises `UnicodeDecodeError` when the payload is not a UTF-8 symbol name (e.g. a binary SUMUP batch), and `get_variable_by_name` raises on an unknown symbol. Catch handler exceptions at the dispatch, log them, and return an ADS device error.

### Testing
- `tests/test_testserver.py` passes (3/3).
- Added-behaviour checks: notification handles are unique across variables; a
  read-write carrying a non-UTF-8 name now returns ADS error `0x0700` instead of killing the connection thread.

### DIFF
--- a/src/pyads/testserver/advanced_handler.py
+++ b/src/pyads/testserver/advanced_handler.py
@@ -163,11 +163,19 @@ class PLCVariable:
         self.value = value
 
     def register_notification(self) -> int:
-        """Register a new notification."""
+        """Register a new notification.
 
-        handle = self.notification_count
+        Notification handles must be unique across all variables: real TwinCAT
+        returns globally-unique handles and clients key their dispatch on them.
+        ``notification_count`` is a class attribute, but ``self.notification_count
+        += 1`` rebinds it as a per-instance value, so the first notification on
+        every variable returns the same handle (10), the second 11, and so on.
+        Increment the shared class counter instead, so handles never collide.
+        """
+
+        handle = PLCVariable.notification_count
+        PLCVariable.notification_count += 1
         self.notifications.append(handle)
-        self.notification_count += 1
         return handle
 
     def unregister_notification(self, handle: int = None):
@@ -486,7 +494,20 @@ class AdvancedHandler(AbstractHandler):
 
         # Try to map the command id to a function, else return error code
         if command_id in function_map:
-            content = function_map[command_id]()
+            try:
+                content = function_map[command_id]()
+            except Exception:
+                # A command handler that raises must not crash this connection's
+                # handler thread: that leaves the client's later requests
+                # unanswered until they time out, and wedges the server. Most
+                # commonly handle_read_write() does write_data.decode() on a
+                # read-write whose payload is not a UTF-8 symbol name (e.g. a
+                # binary SUMUP batch), or get_variable_by_name() is given an
+                # unknown symbol. Return an ADS device error instead.
+                logger.exception(
+                    "Command handler %s raised", hex(command_id))
+                error_code = struct.pack("<I", 0x0700)  # ADSERR_DEVICE_ERROR
+                return AmsResponseData(state, error_code, b"")
 
         else:
             logger.info("Unknown Command: {0}".format(hex(command_id)))
--- a/src/pyads/testserver/advanced_handler.py
+++ b/src/pyads/testserver/advanced_handler.py
@@ -378,12 +378,25 @@ class AdvancedHandler(AbstractHandler):
                 data = write_data[num_requests * 12 :]
                 offset = 0
 
+                # ADS reports one error code per sub-request. Handle each write
+                # independently so a failure in one sub-request neither discards
+                # the writes that already landed nor collapses the whole batch to
+                # a single error. pyads' own adsSumWriteBytes parses exactly this
+                # per-item error array.
+                error_codes = []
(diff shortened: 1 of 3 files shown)


yannS2016:
Added some test for the feature and fixes mentions in the PR

stlehmann (commented):
Thanks for tracking these down — the root causes are all correctly diagnosed (class-vs-instance counter, unguarded `recvfrom`, uncaught handler exception), and the fixes match real TwinCAT behavior. A few smaller things worth a look before merge:

**1. `advanced_handler.py` — `SUMUP_WRITE` (~L357-378) + new catch-all (~L497)**
One bad sub-request in a SUMUP_WRITE batch now returns a single `0x0700` batch error — but earlier sub-writes (and their notification callbacks) already landed before the exception hit. Real ADS reports one error code *per sub-request*: `adsSyncReadWriteReqEx2` already allocates `index_offset * 4` response bytes for exactly this, and `adsSumWriteBytes`/`adsSumRead` already parse it that way. The `SUMUP_WRITE` branch already builds a same-shaped array, just hardcoded to all-zeros — filling it with real per-item codes on a per-iteration catch would match both the protocol and pyads' own client parsing.

**2. `testserver.py` (~L197-202)**
The new `except (ConnectionResetError, OSError)` around `recvfrom` swallows the error with no log line, unlike the dispatch-level catch in `advanced_handler.py` which calls `logger.exception(...)`. A real socket bug would now vanish silently — worth a log line here too.

**3. Minor, same line** — `ConnectionResetError` is already an `OSError` subclass, so the tuple is redundant with just `OSError`.

**4. Minor, `advanced_handler.py` (~L502)** — `0x0700` is correct (1792 = ADS "device error" class, per `errorcodes.py`), just unnamed — pyads has no named error-code constants anywhere yet, and the file's existing "unknown command" line has the same bare-literal style. Worth a follow-up constant, not a blocker.

---
_Generated by [Claude Code](https://claude.ai/code)_

yannS2016:
> Thanks for tracking these down — the root causes are all correctly diagnosed (class-vs-instance counter, unguarded `recvfrom`, uncaught handler exception), and the fixes match real TwinCAT behavior. A few smaller things worth a look before merge:
> 
> **1. `advanced_handler.py` — `SUMUP_WRITE` (~L357-378) + new catch-all (~L497)** One bad sub-request in a SUMUP_WRITE batch now returns a single `0x0700` batch error — but earlier sub-writes (and their notification callbacks) already landed before the exception hit. Real ADS reports one error code _per sub-request_: `adsSyncReadWriteReqEx2` already allocates `index_offset * 4` response bytes for exactly this, and `adsSumWriteBytes`/`adsSumRead` already parse it that way. The `SUMUP_WRITE` branch already builds a same-shaped array, just hardcoded to all-zeros — filling it with real per-item codes on a per-iteration catch would match both the protocol and pyads' own client parsing.
> 
> **2. `testserver.py` (~L197-202)** The new `except (ConnectionResetError, OSError)` around `recvfrom` swallows the error with no log line, unlike the dispatch-level catch in `advanced_handler.py` which calls `logger.exception(...)`. A real socket bug would now vanish silently — worth a log line here too.
> 
> **3. Minor, same line** — `ConnectionResetError` is already an `OSError` subclass, so the tuple is redundant with just `OSError`.
> 
> **4. Minor, `advanced_handler.py` (~L502)** — `0x0700` is correct (1792 = ADS "device error" class, per `errorcodes.py`), just unnamed — pyads has no named error-code constants anywhere yet, and the file's existing "unknown command" line has the same bare-literal style. Worth a follow-up constant, not a blocker.
> 
> _Generated by [Claude Code](https://claude.ai/code)_

Thanks, all addressed in the latest commit:

1. SUMUP_WRITE now handles each sub-write independently and returns one error code per sub-request (0 / 0x0700); successful writes still land, and the array matches adsSumWriteBytes. Added a handler-level regression test.
2. The recvfrom guard now logs via logger.exception like the dispatch catch, so a real socket fault isn't swallowed.
3. Narrowed to `except OSError` (dropped the redundant ConnectionResetError). Left 0x0700 as a bare literal to match the file's existing style.

stlehmann:
Great work, thanks for your contribution @yannS2016

